PR #7, opened 2022-06-01
Updated User Model

hamza-yusuff:
Updated User Model with the roles and branch fields.

## Notion ticket link

[Update User Model] (https://www.notion.so/uwblueprintexecs/Update-user-model-9742b9cab9bb47b9aea46a750f57709f)

## Implementation description
* Added the enum branch field to the user model
* Changed the migration procedure to allow the enum type to be recognised. Made changes to add_user_model.py file at backend/migration/version - had to use alembiac and sqlalchemy.dialects libraries.  Closely followed this link to make the changes - https://stackoverflow.com/questions/37848815/sqlalchemy-postgresql-enum-does-not-create-type-on-db-migrate.

## Steps to test
1.  Access the postgresel database, and run - SELECT * From users; 

* The migration procedure outlined on the project repo could be reviewed, and changed appropriately since the use of enums require manual change of .py migration files generated.


backend/python/migrations/versions/Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,36 @@
+"""empty message
+
+Revision ID: e7f7ce362f62
```
jennchenn, line 3: to make it easier to find the revisions in the future, for now do you want to keep the original filename (i think it should've been something like `e7f7ce362f62_add_driver_role_and_branch_to_user_model.py`)? we can edit the `alembic.ini` file later if we find that there's another format we'd like the files to have (we should also edit the existing versions in that case so they're consistent with whatever new naming convention we choose)

backend/python/migrations/versions/Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,36 @@
...
+    op.add_column(
+        "users", sa.Column("branch", sa.Enum("Algoma", name="branches"), nullable=True)
+    )
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("users", "branches")
```
jennchenn, line 33:
```suggestion
    op.drop_column("users", "branch")
```

jennchenn: If I'm reading your model file correctly, I think the column name should be `branch` (and the enum name is `branches`?)

backend/python/migrations/versions/Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,36 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "e7f7ce362f62"
+down_revision = "2d6c26fde419"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+
```
jennchenn, line 20: it looks like the change to the user didn't get picked up:
```suggestion
     op.execute("ALTER TYPE roles ADD VALUE 'Driver'")
```

backend/python/migrations/versions/Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,36 @@
...
+    branches_stat = postgresql.ENUM("Algoma", name="branches")
+    branches_stat.create(op.get_bind())
```
jennchenn, line 22: if you remove these two lines, does the enum not get created properly? i'm taking a look at the initial migration file since they also add an enum, and it doesn't look like they use `op.get_bind()` https://github.com/uwblueprint/childrens-aid-society/blob/0c7838db69596a1b3d45f7d70f0be8bd6492ee8b/backend/python/migrations/versions/dedc5f7be10c_initial_migration.py#L38

backend/python/migrations/versions/Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("users", "branches")
+    # ### end Alembic commands ###
+    branches_stat = postgresql.ENUM("Algoma", name="branches")
+    branches_stat.drop(op.get_bind())
```
jennchenn, line 36: i think you may be able to change this to:
`op.execute("DROP TYPE branches")`

backend/python/migrations/versions/Add_role_and_branch_to_user_model.py
jennchenn: similarly in downgrade you would need to add code to remove the `Driver`... it may look something like this (taken from https://medium.com/makimo-tech-blog/upgrading-postgresqls-enum-type-with-sqlalchemy-using-alembic-migration-881af1e30abe)
```
op.execute("ALTER TYPE roles RENAME TO roles_old")
op.execute("CREATE TYPE roles AS ENUM('User', 'Admin')")
op.execute((
        "ALTER TABLE users ALTER COLUMN role TYPE roles USING "
        "roles::text::roles"
    ))
op.execute("DROP TYPE roles_old")
```

jennchenn (changes requested):
Nice work on your first PR!!

A few questions/suggestions. Also is the change you made to `db-init/create-multiple-dbs.sh` running `chmod +x`?

backend/python/migrations/versions/Add_role_and_branch_to_user_model.py
hamza-yusuff: Yess sorry about that should have tested the function

backend/python/migrations/versions/Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,36 @@
+"""Add driver role and branch to user model
+
+Revision ID: e7f7ce362f62
+Revises: 2d6c26fde419
```
jennchenn, line 4:
```suggestion
Revises: 797bfedc3a06
```

backend/python/migrations/versions/e7f7ce362f62_Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,36 @@
+"""Add driver role and branch to user model
+
+Revision ID: e7f7ce362f62
+Revises: 2d6c26fde419
```
jennchenn, line 4: nit:
```suggestion
Revises: 797bfedc3a06
```

backend/python/migrations/versions/e7f7ce362f62_Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,36 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute("ALTER TYPE roles ADD VALUE 'Driver'")
+    op.add_column(
+        "users", sa.Column("branch", sa.Enum("Algoma", name="branches"), nullable=True)
+    )
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
```
jennchenn, line 32: I think you'll need to add commands here to revert the changes to the roles enum (see my previous comment) https://github.com/uwblueprint/childrens-aid-society/pull/7#discussion_r886794991

jennchenn (changes requested):
few more changes!

backend/python/migrations/versions/e7f7ce362f62_Add_role_and_branch_to_user_model.py
hamza-yusuff: Sorry about that!

backend/python/migrations/versions/e7f7ce362f62_Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,43 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("users", "branch")
+    # ### end Alembic commands ###
+    op.execute("DROP TYPE branches")
+
+    op.execute("ALTER TYPE roles RENAME TO roles_old")
+    op.execute("CREATE TYPE roles AS ENUM('User', 'Admin')")
+    op.execute(
+        ("ALTER TABLE users ALTER COLUMN role TYPE roles USING " "roles::text::roles")
```
jennchenn, line 41: your command didn't work for me, but this did:
`op.execute("ALTER TABLE users ALTER COLUMN role TYPE roles USING role::text::roles")`

try upgrading/downgrading and see if it works for you

jennchenn (changes requested):

backend/python/migrations/versions/e7f7ce362f62_Add_role_and_branch_to_user_model.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("users", "branch")
+    # ### end Alembic commands ###
+    op.execute("DROP TYPE branches")
+    op.execute("ALTER TABLE users ALTER COLUMN role TYPE roles USING role::text::roles")
```
jennchenn, line 35:
```suggestion
    op.execute("ALTER TYPE roles RENAME TO roles_old")
    op.execute("CREATE TYPE roles AS ENUM('User', 'Admin')")
    op.execute("ALTER TABLE users ALTER COLUMN role TYPE roles USING role::text::roles")
    op.execute("DROP TYPE roles_old")
```

jennchenn (approved):

jennchenn:
lgtm nice work!!! 🚢

PR #19, opened 2022-06-09
created branches model & updated user model

juthikahoque:
## Notion ticket link

[Create Branches Model](https://www.notion.so/uwblueprintexecs/Create-branches-model-8b734166994642a2aa96f122875de528)

## Implementation description
* created a new branches model to match the schema described in the ticket
* updated the user model to reference the id from the branches table as a foreign key
* generated migrations
* updated the upgrade and downgrade functions in migration file & added Algoma as a branch 

## Steps to test
1. flask db upgrade & downgrade should function without errors
2. in the database, there should be a branches table with an entry for Algoma
3. whenever we insert into the users table, an error should pop up if we add an id for the branch_id column that doesn't exist in the branches table

* 

## Checklist


backend/python/migrations/versions/6cd80740889b_reference_user_branch_using_lookup_table.py
```diff
@@ -0,0 +1,38 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "branches",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("branch", sa.String(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.add_column(
```
jennchenn, line 27: getting an error when running upgrade - i think you'll have to:
1. drop the column `branch` that exists on the user table already
2. drop the enum `branches` (`DROP TYPE branches`) 
3. create the `branches` table
4. add the column `branch_id` to the `users` table

... and then do the reverse when downgrading

jennchenn (changes requested):
the upgrade command doesn't seem to be working for me - do you mind taking another look and see if you forgot to commit some changes (i thought we got it mostly working at our last work session)?

also just hold off on rebasing/merging main into your branch for now

backend/python/migrations/versions/6cd80740889b_reference_user_branch_using_lookup_table.py
```diff
@@ -0,0 +1,38 @@
...
+        "users", sa.Column("branch_id", sa.Integer(), sa.ForeignKey("branches.id"))
+    )
+    op.execute("INSERT INTO branches (branch) VALUES ('Algoma')")
```
jennchenn, line 30:
```suggestion
    op.execute("INSERT INTO branches (branch) VALUES ('ALGOMA')")
```

jennchenn: sorry changed my mind about how to store - do you mind just making it all uppercase?

backend/python/app/models/branches.py
```diff
@@ -0,0 +1,25 @@
...
+from sqlalchemy.orm.properties import ColumnProperty
+
+from . import db
+
+
+class Branches(db.Model):
+    __tablename__ = "branches"
+
+    id = db.Column(db.Integer, primary_key=True, nullable=False)
+    branch = db.Column(db.Text(), nullable=False)
```
jennchenn, line 11:
```suggestion
    branch = db.Column(db.String, nullable=False)
```

backend/python/app/models/branches.py
```diff
@@ -0,0 +1,25 @@
+from sqlalchemy import inspect
+from sqlalchemy.orm.properties import ColumnProperty
+
+from . import db
+
+
+class Branches(db.Model):
```
jennchenn, line 7: can you change the class name `Branch` (singular)
+ update the file name (but leave the tablename as `branches`)

backend/python/app/models/user.py
```diff
@@ -15,7 +15,7 @@ class User(db.Model):
     last_name = db.Column(db.String, nullable=False)
     auth_id = db.Column(db.String, nullable=False)
     role = db.Column(roles_enum)
-    branch = db.Column(branches_enum)
+    branch_id = db.Column(db.Integer, db.ForeignKey("branches.id"), nullable=False)
```
jennchenn, line 18: sorry i realize now my slack comment was unclear 🙏 

what i meant to say was can you change the `User.branch_id` back to `User.branch` with the type `db.String` -> we can do validation in the backend when we add a user to ensure the branch is a valid value

jennchenn (commented):
super minor things (mostly my bad for unclear instructions sorry)! 

once you fix these minor things up, the last step will be to rebase off main (lets do this together at tomorrow's work session since we will need to coordinate with some of the other PRs)

backend/python/app/models/__init__.py
```diff
@@ -7,6 +7,7 @@
 
 def init_app(app):
     from .address import Address
+    from .branches import Branches
```
jennchenn, line 10:
```suggestion
    from .branchimport Branches
```

PR #28, opened 2022-06-13
Created services to allow creation of new address

hamza-yusuff:
## Notion ticket link

[Create Service to add new address](https://www.notion.so/uwblueprintexecs/Create-service-to-add-new-address-350083b689fd47e6998d4b65d82d4786)

## Implementation description
* Created class, and the associated method to allow creation of a new address record in the db when a CreateAddressDTO is passed to the method. AddressDTO is returned if the function is successful otherwise exception is raised.
* Added corresponding DTO classes with validation methods.
* Added tests to verify functionality

## Steps to test
1) Upgrade your database docker exec -it cas_py_backend /bin/bash -c "flask db upgrade"
2) Run tests: docker exec -it cas_py_backend /bin/bash -c "pip install -e . && pytest"
3) Verify that all of the tests pass

* Do all the tests pass

## Checklist

### DIFF
--- a/backend/python/app/resources/address_dto.py
+++ b/backend/python/app/resources/address_dto.py
@@ -1,5 +1,5 @@
 class AddressDTO:
-    def __init__(self, id, street_address, city, postal_code, latitude, longitude):
+    def __init__(self, city, id, latitude, longitude, postal_code, street_address):
         self.id = id
         self.city = city
         self.street_address = street_address
@@ -13,19 +13,25 @@ class CreateAddressDTO:
         self.street_address = kwargs.get("street_address")
         self.city = kwargs.get("city")
         self.postal_code = kwargs.get("postal_code")
-        self.latitude = kwargs.get("latitude")
-        self.longitude = kwargs.get("longitude")
+        self.latitude = kwargs.get("latitude", 0.0)
+        self.longitude = kwargs.get("longitude", 0.0)
 
     def validate(self):
         error_list = []
         if type(self.street_address) is not str:
-            error_list.append("The street_address supplied is not a string.")
+            error_list.append("The street_address value supplied is not a string.")
+        if type(self.street_address) is str and self.street_address == "":
+            error_list.append("The street_address value supplied is empty.")
         if type(self.city) is not str:
-            error_list.append("The city supplied is not a string.")
+            error_list.append("The city value supplied is not a string.")
+        if type(self.city) is str and self.city == "":
+            error_list.append("The city value supplied is empty.")
         if type(self.postal_code) is not str:
-            error_list.append("The postal_code supplied is not a string.")
+            error_list.append("The postal_code value supplied is not a string.")
+        if type(self.postal_code) is str and self.postal_code == "":
+            error_list.append("The postal_code value supplied is empty.")
         if type(self.latitude) is not float:
-            error_list.append("The latitude supplied is not a float.")
+            error_list.append("The latitude value supplied is not a float.")
         if type(self.longitude) is not float:
-            error_list.append("The longitude supplied is not a float.")
+            error_list.append("The longitude value supplied is not a float.")
         return error_list
--- a/backend/python/app/resources/address_dto.py
+++ b/backend/python/app/resources/address_dto.py
@@ -1,0 +1,31 @@
+class AddressDTO:
+    def __init__(self, id, street_address, city, postal_code, latitude, longitude):
+        self.id = id
+        self.city = city
+        self.street_address = street_address
+        self.postal_code = postal_code
+        self.latitude = latitude
+        self.longitude = longitude
+
+
+class CreateAddressDTO:
+    def __init__(self, **kwargs):
+        self.street_address = kwargs.get("street_address")
+        self.city = kwargs.get("city")
+        self.postal_code = kwargs.get("postal_code")
+        self.latitude = kwargs.get("latitude")
+        self.longitude = kwargs.get("longitude")
+
+    def validate(self):
+        error_list = []
+        if type(self.street_address) is not str:
+            error_list.append("The street_address supplied is not a string.")
+        if type(self.city) is not str:
+            error_list.append("The city supplied is not a string.")
+        if type(self.postal_code) is not str:
+            error_list.append("The postal_code supplied is not a string.")
+        if type(self.latitude) is not float:
+            error_list.append("The latitude supplied is not a float.")
+        if type(self.longitude) is not float:
+            error_list.append("The longitude supplied is not a float.")
+        return error_list
--- a/backend/python/app/services/implementations/address_service.py
+++ b/backend/python/app/services/implementations/address_service.py
@@ -5,20 +5,21 @@ from ..interfaces.address_service import IAddressService
 
 
 class AddressService(IAddressService):
-
     def __init__(self, logger):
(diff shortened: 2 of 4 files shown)


backend/python/tests/functional/test_address_service.py
```diff
@@ -0,0 +1,79 @@
...
+        # boundary case with city being assigned an empty string
+        CreateAddressDTO(
+            street_address="333 Lester",
+            city="",
+            postal_code="N2L 3W6",
+            latitude=8.380900,
+            longitude=9.899900,
+        ),
+    ]
+    test_address_service = AddressService(current_app.logger)
```
jennchenn, line 43: similar to how it's done in the other files right now, can you move this out into a fixture at the top of the file?
https://github.com/uwblueprint/childrens-aid-society/blob/0258555311c572b1d58c18becba0bb5d4a8d4667/backend/python/tests/functional/test_user_service.py#L24-L28

backend/python/tests/functional/test_address_service.py
```diff
@@ -0,0 +1,79 @@
...
+            street_address="333 Lester",
+            city="Waterloo",
+            postal_code="N2L 3W6",
+            latitude=8.380900,
+            longitude=9.899900,
+        ),
+        # boundary case with city being assigned an empty string
+        CreateAddressDTO(
+            street_address="333 Lester",
+            city="",
```
jennchenn, line 37: this is an interesting case - we actually _don't_ want empty strings for `street_address`, `city`, `postal_code`
your validation currently doesn't handle this, but at the very least i don't think we should include this under the valid cases

backend/python/tests/functional/test_address_service.py
```diff
@@ -0,0 +1,79 @@
+import pytest
+from flask import current_app
+
+from app.resources.address_dto import AddressDTO, CreateAddressDTO
+from app.services.implementations.address_service import AddressService
+
+
+def test_create_address_success():
```
jennchenn, line 8: can you actually separate each of these into their own tests rather than running them in one batch (this applies for both the valid and invalid test cases)?

in the case of errors, it'll make it easier to pinpoint which test case is causing the failure if we have them isolated in their own tests (additionally this separates out the testing so one failed test case doesn't cause all of them to not be run)

as a bonus, if we implement test code coverage reporting, having the test cases separated out will be useful too!

backend/python/tests/functional/test_address_service.py
```diff
@@ -0,0 +1,79 @@
...
+    valid_cases = [
+        # normal case
```
jennchenn, line 10: for the normal case, you could also validate the fields in the return object match the input (not necessary for all the test cases, but maybe just one)

backend/python/tests/functional/test_address_service.py
```diff
@@ -0,0 +1,79 @@
...
+        ),
+        # boundary case with city being assigned an empty string
+        CreateAddressDTO(
+            street_address="333 Lester",
+            city="",
+            postal_code="N2L 3W6",
+            latitude=8.380900,
+            longitude=9.899900,
+        ),
+    ]
```
jennchenn, line 42: since `latitude` and `longitude` are the only non-nullable fields right now, can you add a test case where neither are provided to ensure it works?

backend/python/tests/functional/test_address_service.py
```diff
@@ -0,0 +1,79 @@
...
+    test_address_service = AddressService(current_app.logger)
+    for valid in valid_cases:
+        address_instance = test_address_service.create_address(valid)
+        assert type(address_instance) is AddressDTO
+
+
+def test_create_address_invalid_field_fails():
```
jennchenn, line 49: since you added validation for an empty address being passed to the service, you can add that test case here!

```
with pytest.raises(Exception):
     test_address_service.create_address(None)
```

jennchenn (changes requested):
nice work thinking up some unique test cases!!

added some thoughts + suggestions for potential refactors/testing improvements

backend/python/app/resources/address_dto.py
```diff
@@ -0,0 +1,31 @@
+class AddressDTO:
+    def __init__(self, id, street_address, city, postal_code, latitude, longitude):
+        self.id = id
```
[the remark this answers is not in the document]
hamza-yusuff: Done!

backend/python/app/resources/address_dto.py
```diff
@@ -0,0 +1,31 @@
...
+        self.longitude = longitude
+
+
+class CreateAddressDTO:
+    def __init__(self, **kwargs):
+        self.street_address = kwargs.get("street_address")
+        self.city = kwargs.get("city")
+        self.postal_code = kwargs.get("postal_code")
+        self.latitude = kwargs.get("latitude")
+        self.longitude = kwargs.get("longitude")
```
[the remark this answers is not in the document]
hamza-yusuff: Made the changes!

backend/python/tests/functional/test_address_service.py
hamza-yusuff: Just a bit confused on this. Do i separate the cases into their own function? for example functions like - test_create_address_success_normal_case(), test_create_address_success_boundary_case(), test_create_address_success_precision_case() - for each test case?

backend/python/app/services/implementations/address_service.py
```diff
@@ -0,0 +1,39 @@
...
+from ..interfaces.address_service import IAddressService
+
+
+class AddressService(IAddressService):
+    def __init__(self, logger):
+        self.logger = logger
+
+    def create_address(self, address):
+        try:
+            if address:
```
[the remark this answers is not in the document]
hamza-yusuff: Yess, changed the code to log the error only once, also added checks to see if it is of type CreateDTO or not

backend/python/tests/functional/test_address_service.py
jennchenn: Yes - you can test more than once for normal case, but I think this makes it clearer when we analyze the test results after the tests are run (i.e. instead of seeing `test_create_address_success` fail and having to check which of the specific cases caused it to fail, we can pinpoint that it's a boundary case or precision, etc. that caused the failure)

^ you don't have to separate it out if it seems pretty similar to you, but i think if you have different conditions/things you're testing, would be good to separate it out

backend/python/tests/functional/test_address_service.py
hamza-yusuff: Done!

backend/python/tests/functional/test_address_service.py
hamza-yusuff: I created tow separate class for the normal and invalid cases, and then put the test cases as methods for the classes to keep the cases grouped according to what they test essentially.

backend/python/tests/functional/test_address_service.py
hamza-yusuff: Yess I have put a test case which validates the return object's field with the input's

backend/python/tests/functional/test_address_service.py
hamza-yusuff: Yeah added this well

jennchenn (approved):
i think this looks good!! nice 🥳

PR #31, opened 2022-06-22
Add method get_all_transportation_methods

thrvg:
## Notion ticket link

[Get all transportation methods](https://www.notion.so/uwblueprintexecs/134e1f159caa4f73939d11a700c325ad?v=39cf7d5a98384eb9af2bff2265d0ac8a&p=dabf303df3f842889bfaa96c68ff1f0d)

## Implementation description
* Modified transportation_method_service implementation and interface to add method get_transportation_methods
* Used TransportationMethod.query to get data and modified the data with to_dict() 
* Added error checking
* Added tests 

## Steps to test
1. Run tests
2. Verify code changes in 3 files

* Run the tests. One test should fail (empty db) as I was not sure how to implement that test.

## Checklist

### DIFF
--- a/backend/python/tests/functional/test_transport_method_service.py
+++ b/backend/python/tests/functional/test_transport_method_service.py
@@ -60,23 +60,11 @@ def test_get_transport_method_raises_exception_invalid_argument(
 
 def test_get_transportation_methods_success(transport_method_service):
     res = transport_method_service.get_transportation_methods()
-    # Check if list is returned
     assert type(res) == list
-    # Check if returned list is not empty
-    assert res
-    # Check if returned list has TransportationMethodDTOs
-    assert type(res[0]) is TransportationMethodDTO
-    # Check if correct number of items are returned
     assert len(res) == 3
-    # Check values of all items
-    transportation_methods = set(
-        [key["transportation_method"] for key in DEFAULT_TRANSPORTATION_METHODS]
-    )
-    for transportation_method_object in res:
-        assert (
-            transportation_method_object.transportation_method in transportation_methods
-        )
-        transportation_methods.remove(
-            transportation_method_object.transportation_method
-        )
-    assert len(transportation_methods) == 0
+    assert all(type(item) == TransportationMethodDTO for item in res)
+    transportation_methods_db = [
+        entry["transportation_method"] for entry in DEFAULT_TRANSPORTATION_METHODS
+    ]
+    transportation_methods_res = [item.transportation_method for item in res]
+    assert set(transportation_methods_db) == set(transportation_methods_res)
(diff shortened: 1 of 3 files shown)


jennchenn (changes requested):
looking good so far! we can discuss how to best do the tests

thrvg:
Hi Jenn!

Made the changes. Please let me know if apart from the tests written in `test_get_transportation_method_success(..)`, everything else looks okay. :)

We could discuss the test implementation on Sunday.

backend/python/tests/functional/test_transport_method_service.py
```diff
@@ -48,11 +48,35 @@ def test_get_transport_method_raises_exception_null_argument(
...
+        [key["transportation_method"] for key in DEFAULT_TRANSPORTATION_METHODS]
+    )
+    for transportation_method_object in res:
+        assert (
+            transportation_method_object.transportation_method in transportation_methods
+        )
+        transportation_methods.remove(
+            transportation_method_object.transportation_method
+        )
+    assert len(transportation_methods) == 0
```
jennchenn, line 82: thoughts on something like this instead? it should do the same checks but i think this is a bit more understandable (feel free to modify though!)

also i don't think we need to comment exactly what we're doing unless it's really unclear (e.g. you could add a comment like `Checking return matches expected transportation methods`, but other ones i don't think are necessary)

```suggestion
    assert type(res) == list
    assert len(res) == 3
    assert all(type(item) == TransportationMethodDTO for item in res)

    transportation_methods_db = [
        entry["transportation_method"] for entry in DEFAULT_TRANSPORTATION_METHODS
    ]
    transportation_methods_res = [item.transportation_method for item in res]
    assert set(transportation_methods_db) == set(transportation_methods_res)

```

thrvg: Yes, definitely increases readability! Will commit the changes.

backend/python/tests/functional/test_transport_method_service.py
```diff
@@ -48,11 +48,23 @@ def test_get_transport_method_raises_exception_null_argument(
...
 ):
     with pytest.raises(Exception):
-        transport_method_service._get_transport_method(12345)
+        transport_method_service.get_transportation_method(12345)
+
+
+def test_get_transportation_methods_success(transport_method_service):
+    res = transport_method_service.get_transportation_methods()
+    assert type(res) == list
+    assert len(res) == 3
```
jennchenn, line 64: nit
```suggestion
    assert len(res) == len(DEFAULT_TRANSPORTATION_METHODS)
```

jennchenn (approved):
nice work!

PR #34, opened 2022-06-23
Add date to migration file name

hamza-yusuff:
Updated alembic.inic file to generate migration version file with a sequential naming.

## Notion ticket link

[Update Alembic version naming convention](https://www.notion.so/uwblueprintexecs/Update-Alembic-version-naming-convention-8ca2f3719f284bc49443dbb16aa2971e)

## Implementation description
* Changed the file_template variable in the file alembic.inic such that it creates migration file versions with this naming format - YYYYMMDD-description

## Steps to test
1. Make changes to the schema and run migrations OR downgrade to base migration, then upgrade and finally generate a migration file
2. Verify that the migration file generated has naming of this format -> YYYYMMDD-description

* Does the migration file follow the specified format for it's naming?

## Checklist
- [X] My PR name is descriptive and in imperative tense
- [X] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [X] I have run the appropriate linter(s)
- [X] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

### DIFF
--- a/backend/python/migrations/alembic.ini
+++ b/backend/python/migrations/alembic.ini
@@ -3,7 +3,7 @@
 [alembic]
 # template used to generate migration files
 # file_template = %%(rev)s_%%(slug)s
-file_template = %%(year)d_%%(month).2d_%%(day).2d-%%(rev)s_%%(slug)s
+file_template = %%(year)d-%%(month).2d-%%(day).2d_%%(rev)s_%%(slug)s
 # set to 'true' to run the environment during
 # the 'revision' command, regardless of autogenerate
 # revision_environment = false
--- a/backend/python/migrations/alembic.ini
+++ b/backend/python/migrations/alembic.ini
@@ -3,7 +3,7 @@
 [alembic]
 # template used to generate migration files
 # file_template = %%(rev)s_%%(slug)s
-
+file_template = %%(year)d_%%(month).2d_%%(day).2d-%%(slug)s
 # set to 'true' to run the environment during
 # the 'revision' command, regardless of autogenerate
 # revision_environment = false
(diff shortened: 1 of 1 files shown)


backend/python/migrations/alembic.ini
```diff
@@ -3,7 +3,7 @@
 [alembic]
 # template used to generate migration files
 # file_template = %%(rev)s_%%(slug)s
-
+file_template = %%(year)d_%%(month).2d_%%(day).2d-%%(slug)s
```
jennchenn, line 6: can you add `%%(rev)s` to the name (before the slug maybe)

jennchenn (changes requested):
tested and it works!

nit: can you rename PR title to something a bit more descriptive (e.g. `Add date to migration file name`)?

backend/python/migrations/alembic.ini
hamza-yusuff: Actually, the rev_id was there but then I thought would that cause the migration version files to not be in sequence? as in would that affect the sorting?

jennchenn: If you do `<date>_<rev ID>_<slug>` it should remain ordered

hamza-yusuff: Oh alright, having the changes done that way then.

backend/python/migrations/alembic.ini
```diff
@@ -3,7 +3,7 @@
 [alembic]
 # template used to generate migration files
 # file_template = %%(rev)s_%%(slug)s
-
+file_template = %%(year)d_%%(month).2d_%%(day).2d-%%(rev)s_%%(slug)s
```
jennchenn, line 6:
```suggestion
file_template = %%(year)d-%%(month).2d-%%(day).2d_%%(rev)s_%%(slug)s
```

jennchenn (approved):
one formatting nit; otherwise looks good, nice!

PR #59, opened 2022-07-24
Update get/add methods for concern service

hamza-yusuff:
## Notion ticket link

[Update get/add new familial/child concerns method](https://www.notion.so/uwblueprintexecs/Update-get-add-new-familial-child-concerns-method-980c56eb70bc4447a4c76bba78cd2a7d)

## Implementation description
* Changed implementation pattern for `get` methods to query the single `Concern` table for both `familial` and `child` concerns.  Added optional `type` param for the `get` methods. 
* Updated the `add` method to reference the single `Concern` table. 
* Updated the tests to work properly with the changes.

## Steps to test
1. Run the tests to see if all of them pass.

* Optional `type` param for `get` methods could be made non-optional with only one method for `get`. 

## Checklist
- [X] My PR name is descriptive and in imperative tense
- [X] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [X] I have run the appropriate linter(s)
- [X] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


backend/python/app/services/implementations/concern_service.py
```diff
@@ -9,15 +8,25 @@ class ConcernService(IConcernService):
     def __init__(self, logger):
         self.logger = logger
 
-    def get_familial_concern(self, familial_concern):
+    def get_familial_concern(self, familial_concern, type=None):
```
jennchenn, line 11: since you have two separated functions here, is the `type` parameter necessary?

i think you can either:
1. keep the two separate functions, remove `type`
2. combine into one get function with the `type` param logic as is

backend/python/tests/coverage.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+
+def test_coverage_report():
+    os.system("coverage run -m pytest")
+    os.system("coverage report")
+
+
+test_coverage_report()
```
jennchenn, line 9: this should be removed

in the future when you make new branches, do you mind checking out from main? / can you delete this file locally if it still exists it has been creeping into all of your recent PRs (this and the `.coverage` file)

backend/python/tests/unit/test_concern_service.py
```diff
@@ -1,114 +1,110 @@
...
+
+
+def seed_database():
+    child_concern_instance = [Concern(**data) for data in DEFAULT_CHILD_CONCERNS]
+    familial_concern_instance = [Concern(**data) for data in DEFAULT_FAMILIAL_CONCERNS]
+    db.session.bulk_save_objects(child_concern_instance)
+    db.session.bulk_save_objects(familial_concern_instance)
+
+
+def test_get_familial_concern_id_success(concern_service):
```
jennchenn, line 57: nit ```suggestion
def test_get_familial_concern_success(concern_service):
```

File: backend/python/tests/unit/test_concern_service.py [hunk 4]
```diff
@@ -1,114 +1,110 @@
...
+    res = concern_service.get_familial_concern("FAMILY_CONFLICT")
+    assert type(res) is ConcernDTO
+    assert res.concern == "FAMILY_CONFLICT"
+
+
+def test_get_child_concern_id_success(concern_service):
```
jennchenn, line 63: nit ```suggestion
def test_get_child_concern_success(concern_service):
```

File: backend/python/tests/unit/test_concern_service.py [hunk 5]
```diff
@@ -1,114 +1,110 @@
...
+def test_add_familial_concern_success(concern_service):
+    res = concern_service.add_concern("FAMILIAL_CONCERN", "ABUSE")
+    assert type(res) is ConcernDTO
+    assert Concern.query.get(res.id).concern == "ABUSE"
+
+
+def test_add_child_concern_success(concern_service):
+    res = concern_service.add_concern("CHILD_BEHAVIOUR", "TERROR")
+    assert type(res) is ConcernDTO
+    assert Concern.query.get(res.id).concern == "TERROR"
```
jennchenn, line 90: what's the difference between `test_add_familial_concern` and `test_add_familial_concern_success`?

jennchenn: same for `child_concern` variant

jennchenn (changes requested):
few questions about the current implementation + some smaller nits!

backend/python/app/services/implementations/concern_service.py
hamza-yusuff: Since, the tests are designed without the `type`, I'll move ahead with the first option.

backend/python/tests/coverage.py
hamza-yusuff: Yeah I am sorry about this. This somehow manages to creep into the commit. I'll have it removed permanently.

backend/python/tests/unit/test_concern_service.py
hamza-yusuff: They just test different types of familial and child types. I can have one removed since they don't really add much to the testing. So will keep only the `success` variant for both.

backend/python/app/services/implementations/concern_service.py
```diff
@@ -12,12 +11,15 @@ def __init__(self, logger):
     def get_familial_concern(self, familial_concern):
         try:
             familial_concern_upper = familial_concern.upper()
-            familial_concern_entry = FamilialConcern.query.filter_by(
+            familial_concern_entry = Concern.query.filter_by(
                 concern=familial_concern_upper
```
jennchenn, line 15: right now, it looks like `get_familial_concern` and `get_child_concern` are doing the same things

you should add in a `type=` in the filter to differentiate if you would like to keep this methods separate

backend/python/app/services/interfaces/concern_service.py
```diff
@@ -27,22 +27,12 @@ def get_child_concern(self, child_concern):
         """
         pass
 
-    def add_familial_concern(self, familial_concern):
+    def add_concern(self, type, concern):
         """Creates a new db object and returns the ConcernDTO if adding was successful
 
-        :param familial_concern: string of familial_concern
+        :param concern: string of concern, type: type of concern
```
jennchenn, line 33:
```suggestion
        :param concern: string of concern
        :param type: type of concern
```

backend/python/app/services/interfaces/concern_service.py
```diff
@@ -27,22 +27,12 @@ def get_child_concern(self, child_concern):
...
-        pass
-
-    def add_child_concern(self, child_concern):
-        """Creates a new db object and returns the ConcernDTO if adding was successful
-
-        :param child_concern: string of child_concern
-        :return: ConcernDTO
-        :rtype: ConcernDTO
-        :raises Exception: if child_concern is invalid
+        :raises Exception: if concern is invalid
```
jennchenn, line 36:
```suggestion
        :raises Exception: if concern is invalid/adding to database fails
```
are we actually doing any validation on the input in the add method right now?

backend/python/tests/unit/test_concern_service.py
```diff
@@ -1,114 +1,98 @@
...
+    {"concern": "MEDICAL_ILLNESS_OR_DISABILITY", "type": "CHILD_BEHAVIOUR"},
+)
+# TODO: remove this step when migrations are configured to run against test db
+
+
+def seed_database():
+    child_concern_instance = [Concern(**data) for data in DEFAULT_CHILD_CONCERNS]
+    familial_concern_instance = [Concern(**data) for data in DEFAULT_FAMILIAL_CONCERNS]
+    db.session.bulk_save_objects(child_concern_instance)
+    db.session.bulk_save_objects(familial_concern_instance)
```
jennchenn, line 54: nit: you probably don't need to do this separately anymore since they're all going into the same table but I'm okay with us keeping it like this for now

jennchenn (changes requested):

backend/python/app/services/interfaces/concern_service.py
hamza-yusuff: for now I dont think so. I did not see any validation method inside `ConcernDTO` class.

backend/python/app/services/implementations/concern_service.py
```diff
@@ -9,15 +8,18 @@ class ConcernService(IConcernService):
     def __init__(self, logger):
         self.logger = logger
 
-    def get_familial_concern(self, familial_concern):
+    def get_familial_concern(self, familial_concern, type="FAMILIAL_CONCERN"):
```
jennchenn, line 11: i don't think you should make the `type` a param if this is a function to `get_familial_concern`s only
same thing applies to `get_child_concern` + the `interface` 

i think you'd only want `type` to be a param if this function is `get_concern` in general

jennchenn (approved):
nice! 🚀

PR #63, opened 2022-07-29
Implement formik

juthikahoque:
## Notion ticket link

[Implement Formik](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* wrapped each form with formik within the form files
* onSubmit, we set the different form details using state (which exist in the intake page)

## Steps to test
1. checkout branch
2. check that values are being submitted onSubmit (can add a console.log(values) in each of the onSubmits)

* double check whether or not we want to "submit"/save the information when we click the previous and next buttons

## Checklist


frontend/src/components/intake/ReferralForm.tsx
```diff
@@ -19,131 +21,118 @@ export type ReferralDetails = {
...
+                >
+                  REFERRING WORKER
+                </FormLabel>
+                <Input
+                  id="referringWorker"
+                  name="referringWorker"
+                  type="string"
+                  placeholder="Enter name of worker..."
+                  onChange={handleChange}
+                />
```
jennchenn, line 66: it doesn't seem like this uses the initial values as expected - should we be using formik `Field` instead?
```
<Field
  as={Input}
  id="referringWorker"
  name="referringWorker"
  type="string"
  placeholder="Enter name of worker..."
  onChange={handleChange}
/>
```

frontend/src/components/intake/ReferralForm.tsx
```diff
@@ -19,131 +21,118 @@ export type ReferralDetails = {
...
 }: ReferralFormProps): React.ReactElement => {
+  const initialValues: ReferralDetails = {
+    referringWorker: "",
+    referringWorkerContact: "",
+    familyName: "",
+    referralDate: "",
+    cpinFileNumber: "",
+    cpinFileType: "",
+    phoneNumber: "",
+  };
```
jennchenn, line 40: does doing this reset the initial values each time the form is rendered? do we not need to pass the initial values into each form from the intake page?

frontend/src/components/intake/PermittedIndividualsForm.tsx
```diff
@@ -17,93 +19,100 @@ export type PermittedIndividualDetails = {
 };
 
 type PermittedIndividualFormProps = {
-  permittedIndividualDetails: PermittedIndividualDetails;
-  setPermittedIndividualDetails: React.Dispatch<
-    React.SetStateAction<PermittedIndividualDetails>
+  allOtherPermittedIndividuals: PermittedIndividualDetails[];
+  setAllOtherPermittedIndividuals: React.Dispatch<
+    React.SetStateAction<PermittedIndividualDetails[]>
```
jennchenn, line 24: are these residual changes from your previous ticket? i don't think introducing `Formik` should introduce changes here

frontend/src/components/intake/CourtInformationForm.tsx
```diff
@@ -17,103 +17,122 @@ export type CourtDetails = {
...
+            }}
+          >
+            Previous Button
+          </Button>
+          <Button
+            onClick={() => {
+              handleSubmit();
+              nextStep();
+            }}
+          >
```
jennchenn, line 130: in the `ReferralForm`, it looks like you do something like this:
`<Button type="submit">Next Button</Button>`

is there a reason why this form is different? i think aiming to keep things consistent would make it easier to understand in the future

jennchenn (changes requested):
left some general comments (i only commented on a few spots but the comments apply to any similar implementations)!

frontend/src/components/intake/ReferralForm.tsx
juthikahoque: i think that it's almost like the useState for initial values where the initial values for the form are only set on the first load

jennchenn (approved):
I think this looks good! nice work

PR #67, opened 2022-08-07
add input styling

juthikahoque:
## Notion ticket link

[Setup Input Styling](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* added a new input style file and followed the figma to match styling

## Steps to test
1. pull from the branch
2. use the input at / to see if it matches figma designs
Screen Shot 2022-08-07 at 6 32 14 PM

* checking if the styling is correct

## Checklist


frontend/src/theme/inputStyles.tsx
```diff
@@ -0,0 +1,43 @@
...
+        borderRadius: "4px",
+        _placeholder: {
+          color: "gray.600",
+        },
+        _hover: {
+          background: "gray.100",
+        },
+        _focus: {
+          background: "#gray.100",
+          border: "1px solid #5D5E67",
```
jennchenn, line 28: can we use the colors we defined in the color theme file instead of raw hex codes? i think this maps to `gray.600`

frontend/src/theme/inputStyles.tsx
```diff
@@ -0,0 +1,43 @@
...
+        },
+        _invalid: {
+          background: "red.50",
+          border: "1px solid #FF5449",
```
jennchenn, line 32: should be `red.400`

jennchenn (commented):
two small nits

have you tried adding an icon to the input fields as well?

frontend/src/theme/inputStyles.tsx
juthikahoque: i don't think the codes work with the border :((

jennchenn (approved):
nice! will make a new ticket for input+icon styling

frontend/src/theme/inputStyles.tsx
jennchenn: ah makes sense, the current implementation is okay then!

PR #68, opened 2022-08-08
Add default flag to lookup tables

thrvg:
## Notion ticket link

[Add default flag to all lookup tables](https://www.notion.so/uwblueprintexecs/134e1f159caa4f73939d11a700c325ad?v=39cf7d5a98384eb9af2bff2265d0ac8a&p=612128b08cc9432c84a10f4175825216&pm=s)

## Implementation description
* Added default column to 4 tables - goals, concerns, branches, transportation_methods
* Ran migrations
* Updated tests and DTOs to reflect schema change

## Steps to test
1. Run tests normally

* N/A

## Checklist


backend/python/app/models/branch.py
```diff
@@ -7,3 +7,4 @@ class Branch(db.Model, BaseMixin):
 
     id = db.Column(db.Integer, primary_key=True, nullable=False)
     branch = db.Column(db.String, nullable=False)
+    show_by_default = db.Column(db.Boolean, nullable=False, default=False)
```
jennchenn, line 10: i think better behavior is to allow someone to add a row without specifying the `default` flag value--if it is not specified then it gets set to `False`

right now an insert fails because of the not null constraint; i think you may have to change `nullable=True` for the default to apply (or if you can research a bit to see if we can keep the not null constraint but still use the default that'd be great)

backend/python/migrations/versions/2022-08-04_ee20e81724a5_added_default_column.py
```diff
@@ -0,0 +1,44 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("branches", sa.Column("show_by_default", sa.Boolean(), nullable=True))
+    op.add_column("concerns", sa.Column("show_by_default", sa.Boolean(), nullable=True))
+    op.add_column("goals", sa.Column("show_by_default", sa.Boolean(), nullable=True))
+    op.add_column(
+        "transportation_methods",
+        sa.Column("show_by_default", sa.Boolean(), nullable=True),
+    )
+    op.execute("UPDATE branches SET show_by_default = false")
```
jennchenn, line 27: can you make the default here `True` instead?

jennchenn (commented):
if it's not too much work, do you mind adding the default flag to `access_type` as well?

backend/python/app/models/branch.py
thrvg: > we can keep the not null constraint but still use the default
I think that's how it behaves right now! Inserting should work as normal in case of a NULL it uses the default value.

backend/python/migrations/versions/2022-08-04_ee20e81724a5_added_default_column.py
thrvg: Done!

backend/python/app/models/branch.py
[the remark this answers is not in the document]
thrvg: Done!

backend/python/tests/functional/test_concern_service.py
```diff
@@ -19,83 +19,83 @@ def concern_service():
     {
         "concern": "FAMILY_CONFLICT",
         "type": "FAMILIAL_CONCERN",
-        "show_by_default": False,
+        "is_default": False,
```
jennchenn, line 22: nit: not a big deal but `is_default` for all the test seed data can probably be set to `True`

jennchenn (approved):
lgtm!!

PR #73, opened 2022-08-18
create custom input and change all inputs to custom

juthikahoque:
## Notion ticket link

[Create Custom Input Component](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* created a custom input component under common components
* added forwardRef so that we can pass in the ref prop needed for attaching a new file in CourtDetailsForm.tsx

## Steps to test
1. checkout branch
2. check homepage for input with an icon and intake page for input without an icon

Screen Shot 2022-08-17 at 9 24 37 PM

* if the component is placed in the right spot
* if the component looks good -> i had to disable eslint to pass in the props using the spread operator (since that's the easiest way to pass props into the Input component without making anything mandatory/missing any props)

## Checklist


frontend/src/components/common/CustomInput.tsx
```diff
@@ -0,0 +1,32 @@
+import { InputGroup, Input, InputLeftElement } from "@chakra-ui/react";
+import React from "react";
+
+type CustomInputProps = {
```
jennchenn, line 4: typescript allows you to compose a type as as an intersection of types (more info [here](https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types))

have you tried importing `ImportProps` here like
`type CustomInputProps = ImportProps & { ....`

i don't think this'll solve the eslint issue, but might help scope down the props we accept?

frontend/src/components/common/CustomInput.tsx
```diff
@@ -0,0 +1,32 @@
...
+  icon?: JSX.Element;
+  [x: string]: unknown;
```
jennchenn, line 6: can you explain what this line is for?

frontend/src/components/common/CustomInput.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+const CustomInput = React.forwardRef(
+  (
+    { icon, ...props }: CustomInputProps,
+    ref: React.LegacyRef<HTMLInputElement> | undefined,
+  ) => {
+    return (
+      <InputGroup>
+        {icon ? (
```
jennchenn, line 16: can this be simplified with [conditional rendering](https://reactjs.org/docs/conditional-rendering.html)?
i.e.
```
{icon && 
     <InputLeftElement pointerEvents="none">{icon}</InputLeftElement>
}

jennchenn (changes requested):
thoughts on removing the `ref` and using the normal `Input` component for court order referral attachment? it doesn't really fall into the regular text input category, and that'll help simplify the `CustomInput` component code

frontend/src/components/common/CustomInput.tsx
juthikahoque: this is for the ...props paramater

jennchenn (approved):
nice!

PR #79, opened 2022-09-18
Fix login

kabiroberai:
## Notion ticket link
N/A

## Implementation description
* This PR fixes login, which was failing because AuthDTO didn't know about UserDTO's `branch` field.

## Steps to test
1. Try creating an account and logging in (`/login`) before and after this PR.
2. Observe that there's a 500 error prior to applying this patch.

* test_auth_service: This is my first time writing a pytest-based unit test so feedback would be great!

## Checklist

### DIFF
--- a/backend/python/tests/functional/test_auth_service.py
+++ b/backend/python/tests/functional/test_auth_service.py
@@ -1,0 +1,55 @@
+import pytest
+from flask import current_app
+
+from app.models import db
+from app.models.user import User
+from app.services.implementations.auth_service import AuthService
+from app.services.implementations.user_service import UserService
+
+@pytest.fixture(scope="module", autouse=True)
+def setup(module_mocker):
+    module_mocker.patch("app.utilities.firebase_rest_client.FirebaseRestClient.sign_in_with_password", return_value=FirebaseToken())
+    module_mocker.patch("firebase_admin.auth.get_user_by_email", return_value=FirebaseUser())
+
+@pytest.fixture
+def auth_service():
+    user_service = UserService(current_app.logger)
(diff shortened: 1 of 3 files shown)


backend/python/tests/functional/test_auth_service.py
```diff
@@ -0,0 +1,67 @@
...
+    {
+        "auth_id": "A",
+        "first_name": "Jane",
+        "last_name": "Doe",
+        "role": "Admin",
+    },
+)
+
+
+def insert_users():
```
kabiroberai, line 59: this is taken from test_user_service.py; I wonder whether we should unify the two methods or whether that's taking DRY too far.

backend/python/tests/functional/test_auth_service.py
```diff
@@ -0,0 +1,67 @@
...
+    user_instances = [User(**data) for data in TEST_USERS]
+    db.session.bulk_save_objects(user_instances)
+    db.session.commit()
+
+
+def test_login(auth_service):
+    insert_users()
+    auth_service.generate_token("test@test.com", "password")
```
kabiroberai, line 67: No asserts here but login failures throw an exception, and this was failing as expected before applying the auth_dto patch.

jennchenn (approved):
lgtm thanks for catching this

let's leave the tests as is for now, they can be refactored in the future if necessary

backend/python/tests/functional/test_auth_service.py
jennchenn: so is the test here just that these operations do not throw an exception?

kabiroberai: Pretty much, yep

PR #80, opened 2022-09-19
Create stepper

kabiroberai:
## Notion ticket link
[Create intake form stepper component](https://www.notion.so/uwblueprintexecs/Create-intake-form-stepper-component-90e87b8bad9e48a684bbf3c985e2291c)

## Implementation description
* `Stepper` is a React component, which takes a list of pages (just strings for now) + an active page index
* The stepper is laid out using the CSS grid API, though this can be replaced with flexbox if compatibility with super old browsers is a concern; let me know if so.

## Steps to test
1. Navigate to <frontend>/intake
2. Go back and forth between pages, resize, etc

* The margins and padding currently use hardcoded pixel values, since that's what most other components appear to do as well. This is just food for thought atm, but we might want to switch to em/rem eventually for better responsiveness.
* The UI breaks on mobile devices but this seems to be okay per the spec. If this requirement has changed, I can update the PR to improve the mobile layout.

## Checklist

https://user-images.githubusercontent.com/12706786/190971336-0823d048-dcd8-4f46-ba71-087f67e3e877.mov


frontend/src/components/pages/IntakePage.tsx
```diff
@@ -44,57 +45,52 @@ const Intake = (): React.ReactElement => {
...
+        primaryTitle="Initiate New Case"
+        secondaryTitle="Case Management"
+      />
+      <Box textAlign="center" padding="30px 0 40px 0">
+        <Stepper
+          pages={[
+            "Case referral",
+            "Court information",
+            "Individual details",
+            "Program details",
```
kabiroberai, line 86: I don't think we have the program details page yet fwiw – in the UI this simply materializes as there being no way to navigate to that page; the stepper handles it just fine.

kabiroberai:
updated the reviewers since it seems Atharva jumped ship, pun entirely intended

frontend/src/components/intake/Stepper.tsx
```diff
@@ -0,0 +1,97 @@
...
+                  />
+                  <GridItem gridRowStart={2} />
+                </>
+              )}
+              <GridItem
+                gridRowStart={1}
+                justifySelf="center"
+                sx={{ "svg path": { stroke: color } }}
+              >
+                {isComplete ? (
```
jennchenn, line 48: for the check icon, do you mind using [this](https://feathericons.com/?query=check-circle) instead

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -44,57 +45,52 @@ const Intake = (): React.ReactElement => {
...
+          ]}
+          activePage={step - 1}
```
jennchenn, line 88: do you think it makes sense to keep the stepper 0-indexed despite having to do this conversion here to get the `activePage`?

jennchenn (commented):
looking good!!

frontend/src/components/pages/IntakePage.tsx
kabiroberai: I don't think the `Stepper` component should be dependent on the implementation details of a client; in fact I'd suggest making `step` zero-indexed too but that's a separate discussion.

frontend/src/components/intake/Stepper.tsx
kabiroberai: neat! done.

jennchenn (approved):
looks good to me! nice work

PR #82, opened 2022-09-24
Implement "Add Individual" functionality

kabiroberai:
## Notion ticket link

[Support ‘add’ functionality in other permitted individuals screen](https://www.notion.so/uwblueprintexecs/Support-add-functionality-in-other-permitted-individuals-screen-94b3856f929d4ac2b79c51ca9e02d2b1)

## Implementation description
* I decided to use a single form rather than creating multiple instances of PermittedIndividualsForm, since doing it this way plays a bit better with Formik's data flow arch
* [Formik.FieldArray](https://formik.org/docs/api/fieldarray) is well suited for this purpose.

## Steps to test
1. Go to the Other Permitted Individuals page of `/intake`
2. Add and remove people to your heart's content

- I would've attempted to base this off of fcb6c9e but I only realized that Juthika's branch existed about halfway through my implementation; reconciling the two approaches likely wouldn't afford us much at this point, since they diverge pretty significantly.
- Formik doesn't call `setPermittedIndividualDetails` until the user submits. Consequently, I'm not sure whether modelling permittedIndividualDetails (and by extension any other form data) as `IntakePage` "state" is the correct abstraction, at least not now. We should either 1) make the state update live (using `Formik.useField` or similar), or 2) clearly document the current behavior as expected. ~~I'm leaning towards the former since it allows for live validation, autosave across page loads, etc.~~ EDIT: After reading Formik's docs more thoroughly, I realize it might just be better if we go with option 2, and rely on `useFormikContext` for autosave and co. I'm curious what you folks think.

Screen Shot 2022-09-24 at 1 45 26 AM
Screen Shot 2022-09-24 at 1 45 20 AM

## Checklist

### DIFF
--- a/frontend/src/components/intake/PermittedIndividualsForm.tsx
+++ b/frontend/src/components/intake/PermittedIndividualsForm.tsx
@@ -8,15 +8,26 @@ import {
   Select,
   SimpleGrid,
 } from "@chakra-ui/react";
-import { Field, Form, Formik } from "formik";
+import { Field, FieldArray, Form, Formik } from "formik";
 import CustomInput from "../common/CustomInput";
 
-export type PermittedIndividualDetails = {
+export type PermittedIndividual = {
   firstName: string;
   lastName: string;
   phoneNumber: string;
   relationship: string;
 };
+
+export type PermittedIndividualDetails = {
+  individuals: PermittedIndividual[];
+};
+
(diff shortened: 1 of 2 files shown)


kabiroberai:
punting this pending further scoping of the Individual Details screen, per @jennchenn

bowenzhu1:
closing in favor of #139  and new tickets

PR #87, opened 2022-09-28
[kai/db-seed-script] Set up database seed script

kai-wei-mo:
## Notion ticket link

[Set up database seed script](https://www.notion.so/uwblueprintexecs/134e1f159caa4f73939d11a700c325ad?v=39cf7d5a98384eb9af2bff2265d0ac8a&p=a541358c29f44c6d86caf516088872e1&pm=s)

## Implementation description
* Add script to populate dev db with mock data
* Add db nuke script for convenience

## Steps to test
1. `docker-compose down --volumes`
2. `docker-compose up`
3. `docker exec -it cas_py_backend /bin/bash -c "python -m tools.db_seed"`
4. `docker exec -it cas_db /bin/bash`
5. `psql -U postgres -d cas`
6. `\d`
7. `select * from children` and other tables as necessary
8. `docker exec -it cas_py_backend /bin/bash -c "python -m tools.db_seed nuke"`
9. enter the db again and see what is there

- cleanliness of code in `backend/python/tools/db_seed.py` (i am new to sqlalchemy)
- readability of code (i.e. 80 char line limit)

## Checklist


backend/python/tools/db_seed.py
```diff
@@ -0,0 +1,294 @@
...
+    db.engine.execute(
+        f"INSERT INTO {table_name} ({tup_to_string_commas(column_names)}) VALUES ({tup_to_string_commas_and_quotes(values)});"
+    )
+
+
+# fmt: off
+def insert_test_data():
+    # Users
+    create_enum(db, "user_roles_enum", ('User', 'Admin'))
+    create_table(db, "users", """
```
kabiroberai, line 61: I don't think it's necessary to run any CREATE commands; the structure should already be defined if you apply all the migrations (`flask db upgrade`). Accordingly, it shouldn't be necessary to rename the enums in `app/models` either.

backend/python/tools/db_seed.py
```diff
@@ -0,0 +1,294 @@
...
+        f"CREATE TYPE {enum_name} AS ENUM ({tup_to_string_commas_and_quotes(enum_values)});"
+    )
+
+
+def insert_values(db, table_name: str, column_names: tuple, values: tuple):
+    """
+    Insert values into a table.
+    """
+    db.engine.execute(
+        f"INSERT INTO {table_name} ({tup_to_string_commas(column_names)}) VALUES ({tup_to_string_commas_and_quotes(values)});"
```
kabiroberai, line 53: IIUC this is a raw SQL statement – I'd avoid using those since it's easy to miss escape sequences, and if parts of this code were ever used in prod it could open up the possibility for SQL injection. Try using the ORM if possible, and if you need raw SQL then maybe go with SQLAlchemy's [SQL Expression Language](https://docs.sqlalchemy.org/en/13/core/tutorial.html) instead.

jennchenn: since this script will only be used as a dev tool, i think it's fine to leave as is (but otherwise agree!)

backend/python/tools/db_seed.py
```diff
@@ -0,0 +1,294 @@
...
+
+    for value in values:
+        insert_values(db, "goals", ("id", "type", "goal", "start_date", "end_date", "is_default"), value)
+
+    # Intake
+    create_enum(db, "court_status_enum", ('INTERIM_CARE', 'FINAL_ORDER_FOR_SOCIETY_CARE', 'EXTENDED_SOCIETY_CARE', 'SUPERVISION_ORDER', 'KIN_SERVICE_PLACEMENT', 'LIVING_WITH_BIO_FAMILY', 'OTHER'))
+    create_enum(db, "first_nation_heritage_enum", ('FIRST_NATION_REGISTERED', 'ELIGIBLE_FOR_REGISTRATION', 'INUIT', 'METIS', 'UNKNOWN'))
+    create_enum(db, "intakes_access_weekday_enum", ('MONDAY', 'TUESDAY', 'WEDNESDAY', 'THURSDAY', 'FRIDAY', 'SATURDAY', 'SUNDAY'))
+
+    # create_table(db, "intake_concerns", """
```
jennchenn, line 156: are these commented out on purpose?

backend/python/tools/db_seed.py
```diff
@@ -0,0 +1,294 @@
...
+def insert_values(db, table_name: str, column_names: tuple, values: tuple):
+    """
+    Insert values into a table.
+    """
+    db.engine.execute(
+        f"INSERT INTO {table_name} ({tup_to_string_commas(column_names)}) VALUES ({tup_to_string_commas_and_quotes(values)});"
+    )
+
+
+# fmt: off
```
jennchenn, line 57: what is this comment for?

backend/python/tools/db_seed.py
```diff
@@ -0,0 +1,294 @@
+import os
+import sys
+from ctypes import addressof
```
jennchenn, line 3: where is this import being used?

backend/python/tools/db_seed.py
```diff
@@ -0,0 +1,294 @@
...
+    db.engine.execute("GRANT ALL ON SCHEMA public TO postgres;")
+    db.engine.execute("GRANT ALL ON SCHEMA public TO public;")
+    db.engine.execute("COMMENT ON SCHEMA public IS 'standard public schema';")
+
+
+if __name__ == "__main__":
+    app = create_app("development")
+    if "nuke" in sys.argv:
+        nuke_db()
+        print(type(db))
```
jennchenn, line 292: i think we can remove this print statement

jennchenn (changes requested):
followed your test steps and things seem to be working!

a few comments: 
1. Since we have migrations setup already, we should be using those to create the tables instead of creating the tables in this script -- this script should simply add the test data. If you'd like to make a wrapper script to both run the migrations and then insert the data, that could be useful (otherwise that can be on the developer to do).

2. The above comment would apply to the "nuke"-ing as well - it would simply remove the test data inserted in the table, but not remove any existing tables. Maybe "clear_data" would make more sense in that case?

3. Why did you choose to make a separate `insert_values` method rather than just having all the values inline with the statement (e.g. [here](https://github.com/uwblueprint/planet-read/blob/247a2c939c3e06cd9822062e497dfc9633e1c23b/backend/python/tools/insert_test_stories.py#L71-L77))? Even though we'd be repeating `INSERT INTO {table_name} {columns} VALUES {values}` for each table, we could then remove the `tup_to_string_commas` and `tup_to_string_commas_and_quotes` methods

backend/python/tools/db_seed.py
kai-wei-mo: mb, this was straight copypasta from another file

backend/python/tools/db_seed.py
kai-wei-mo: suggestion accepted

backend/python/tools/db_seed.py
```diff
@@ -0,0 +1,133 @@
...
+    # Goal
+    # values = [
+    #     (1, "LONG_TERM", "Eat more vegetables",     "2020-01-05", "2023-05-29", False),
+    #     (2, "LONG_TERM", "Floss",                   "2020-02-15", "2023-06-30", True),
+    #     (3, "SHORT_TERM", "Learn to cook for one",  "2020-02-24", "2023-05-30", True),
+    #     (4, "SHORT_TERM", "One Amazing Cartwheel!", "2022-03-06", "2023-06-05", True),
+    # ]
+
+    # for value in values:
+    #     insert_values(db, "goals", ("id", "type", "goal", "start_date", "end_date", "is_default"), value)
```
kai-wei-mo, line 82: i commented this out because `flask db upgrade` already creates these tables with some values. should i delete these lines and go with the values that `flask db upgrade` populates?

backend/python/tools/db_seed.py
kai-wei-mo: this disables linting and formatting tools between `#fmt: off` and `#fmt: on`. i found that the linting made things less readable in this occassion

backend/python/tools/db_seed.py
jennchenn: yup you can delete these lines !

backend/python/app/models/caregiver.py
```diff
@@ -1,7 +1,7 @@
 from . import db
 from .base_mixin import BaseMixin
 
-type_enum = db.Enum("CAREGIVER", "PROVIDER", name="caregivers_type")
+caregiver_type_enum = db.Enum("CAREGIVER", "PROVIDER", name="caregivers_type")
```
kai-wei-mo, line 4: possible source of ci failing?

kai-wei-mo: nvm

backend/python/tests/functional/test_caregiver_service.py
```diff
@@ -26,7 +26,7 @@ def caregiver_service():
     "first_name": "Hamza",
     "last_name": "Yusuff",
     "auth_id": "hbyusuff",
-    "role": "Driver",
+    "role": "User",
```
kai-wei-mo, line 29: changed `Driver` to `User` since i changed `Driver` to no longer be a valid `user_role_enum` in `backend/python/app/models/user.py`. same with `backend/python/tests/functional/test_child_service.py`

backend/python/tools/db_seed.py
kai-wei-mo: noted; i now know to use orm in the future. thanks guys :D

jennchenn (approved):
looks good to me! nice work

PR #88, opened 2022-10-02
Convert Court Information page to hi-fi

kabiroberai:
## Notion ticket link

[Convert Court Information page to hi-fi](https://www.notion.so/uwblueprintexecs/Convert-Court-Information-page-to-hi-fi-08e57388560b4f6dba1d46ff930303a1)

## Implementation description
* Introduce new `Autocomplete*` components built on top of Formik + Chakra's Popover
* Bring court info page up to spec

## Steps to test
Go to the Court Information page of `/intake`.

* One design decision worth discussing: I opted to use `<select>` rather than our custom popover for the First Nation Heritage dropdown, since afaict it has the exact semantics of `<select>` in that you can only choose from one of the pre-defined options. Using the intended component here gives us better accessibility + UX out of the box, but I can create a new select-like autocomplete component if it's important that the design be to-spec.

## Checklist

https://user-images.githubusercontent.com/12706786/193434767-67680ee5-d3ec-4219-9442-47e6cf40d884.mov


kabiroberai:
This PR also tackles some bits of this ticket now since the changes were relatively interlinked: [Make minor cosmetic changes to intake hi-fi](https://www.notion.so/uwblueprintexecs/Make-minor-cosmetic-changes-to-intake-hi-fi-7fd548579ecb464690e588007026e50b)

frontend/src/components/intake/Autocomplete.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+export const AutocompleteSuggestions = ({
+  hints,
+  onSelect,
+}: {
+  hints: string[];
+  onSelect: (value: string) => void;
+}): React.ReactElement => {
+  return (
+    <PopoverContent
```
jennchenn, line 31: how much work would it be to use the `CustomTag` item that @juliannejorda setup? (you can of course make changes to it)

thought behind this is that separating it out will make it easier to reuse for dropdowns as well, but curious what your thoughts are on this

https://github.com/uwblueprint/childrens-aid-society/blob/main/frontend/src/components/common/CustomTag.tsx

frontend/src/components/intake/Autocomplete.tsx
```diff
@@ -0,0 +1,140 @@
...
+      {...{ ...field, ...props }}
+      name={name}
+      variant="filled"
+      height="48px"
+      backgroundColor="gray.50"
+      borderColor="gray.100"
+      borderWidth="1px"
+      fontSize="18px"
+      color={field.value ? "black" : "gray.600"}
+      sx={{ ":hover, :focus": { backgroundColor: "gray.100" } }}
```
jennchenn, line 135: can we make this a new variant for this and put it here https://github.com/uwblueprint/childrens-aid-society/tree/main/frontend/src/theme

frontend/src/components/intake/Autocomplete.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { useField } from "formik";
+import CustomInput, { CustomInputProps } from "../common/CustomInput";
+
+export type AutocompleteFieldProps = CustomInputProps & {
+  name: string;
+  hints: string[];
+};
+
+export const AutocompleteSuggestions = ({
+  hints,
```
jennchenn, line 24: nit: i think `hints` can be renamed -> `suggestions` / `options` / anything else you think is more descriptive

frontend/src/components/intake/CourtInformationForm.tsx
```diff
@@ -78,54 +90,89 @@ const CourtInformationForm = ({
...
+                  id="documentDisplay"
+                  placeholder="No document attached"
+                  value={values.orderReferral?.name}
+                  onClick={handleClick}
+                  marginRight="10px"
+                  style={{ cursor: "pointer" }}
+                />
+                <Button
+                  onClick={handleClick}
+                  variant="ghost"
```
jennchenn, line 114: change to `variant="tertiary"` 

https://github.com/uwblueprint/childrens-aid-society/blob/48f264af53cfca66d811553e212dc8b4065764fb/frontend/src/theme/buttonStyles.tsx#L58-L70

frontend/src/components/intake/Autocomplete.tsx
```diff
@@ -0,0 +1,140 @@
...
+      />
+    </Popover>
+  );
+};
+
+export type SelectFieldProps = SelectProps & {
+  name: string;
+};
+
+export const CustomSelectField = ({
```
jennchenn, line 117: can we move this into its own file?

frontend/src/components/intake/Autocomplete.tsx
```diff
@@ -0,0 +1,140 @@
...
+              {hint}
+            </Box>
+          </Button>
+        ))}
+      </PopoverBody>
+    </PopoverContent>
+  );
+};
+
+export const AutocompleteField = ({
```
jennchenn, line 79: nit: not sure if `AutocompleteField` makes the most sense to me - thoughts on renaming to something like `CustomSelectionField` (and then `CustomSelect` can be renamed to something like `CustomDropdown`?

frontend/src/components/intake/Autocomplete.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+export const CustomSelectField = ({
+  name,
+  children,
+  ...props
+}: SelectFieldProps): React.ReactElement => {
+  const [field] = useField<string>(name);
+
+  return (
+    <Select
```
jennchenn, line 125: can you do a bit of investigation into the effort required to convert this into the custom dropdown matching the figma specs + to keep keyboard/screen reader navigation? you don't have to actually do the work in this branch (i will make a ticket for creating multi-select input fields anyway), but curious to know if it's a big/small lift

frontend/src/components/intake/Autocomplete.tsx
```diff
@@ -0,0 +1,140 @@
...
+    <PopoverContent
+      bg="gray.100"
+      borderColor="gray.50"
+      shadow="none"
+      maxHeight="300px"
+      overflowY="auto"
+      display={hints.length === 0 ? "none" : "block"}
+    >
+      <PopoverBody padding="8px">
+        {hints.map((hint, index) => (
```
jennchenn, line 40: ideally if an item is selected an a user clicks on the field again, i think it should show the full list of options + an [active SelectionItemContainer](https://www.figma.com/file/Pi7j5EmcROKxGE3eUFGhDK/0.-Libraries-%3E-CAS-Design-system?node-id=746%3A10429) to indicate the current selection rather than filtering everything out

^ but let me know if this requires a lot of rework

jennchenn (changes requested):
this is looking really good! some comments/questions 

can you just confirm all changes requested in `Make minor cosmetic changes to intake hi-fi` should be in this branch? i haven't commented on it yet just in case you were planning to separate some of it out

frontend/src/components/intake/CourtInformationForm.tsx
```diff
@@ -23,6 +26,12 @@ type CourtInformationFormProps = {
   prevStep: () => void;
 };
 
+const optional = (
+  <Box display="inline" color="gray.600">
+    (OPTIONAL)
+  </Box>
+);
+
```
jennchenn, line 34: since we will have more than one optional field, do you think it'd make sense to extract this out into another component (e.g. an `OptionalFormLabel` component that'll just add this optional tag after the form label?)

frontend/src/components/intake/Autocomplete.tsx
kabiroberai: The semantic distinction between "Select" and "Autocomplete" is that the latter allows users to enter values that don't match the suggestions, whereas the former has to have a value that's one of the options. That's the same reason I've used the term `hints` instead of `options`, but if you have a strong pref I can still rename this.

frontend/src/components/intake/Autocomplete.tsx
kabiroberai: Getting an MVP working would probably be a 2-story-points kinda task, though keyboard navigation/a11y would definitely require greater eng lift. Still worth doing imo, I'd estimate a couple of hours to get it in a polished state.

frontend/src/components/intake/Autocomplete.tsx
kabiroberai: Done, I also made the component optionally controllable so that clients can read/write the `pressed` prop.

jennchenn (approved):
going to approve this for now so we can get it merged in, thanks for all the great work!!!

do you mind going through and adding `# TODO` comments to the changes you did not address--i will document those in a ticket after

PR #95, opened 2022-10-26
Created footer

Het7:
## Notion ticket link
[Create intake form footer navigation component](https://www.notion.so/uwblueprintexecs/134e1f159caa4f73939d11a700c325ad?v=39cf7d5a98384eb9af2bff2265d0ac8a&p=3e8a3cb2f18742fa91e7f69326a78bdc&pm=s)

## Implementation description
* Created a new component
Screen Shot 2022-10-25 at 9 23 52 PM

## Checklist


jennchenn (changes requested):
talked about it during the meeting but putting this here for transparency:
* please update from main
* the footer should be replacing the prev/next buttons at the bottom of the page
 
please request re-review when addressed!!

Het7:
> 

Addressed!

jennchenn (changes requested):
1. the footer doesn't seem to always stick to the bottom of the page (see screenshot)
image
2. the buttons don't seem to be aligned to the right when the page is resized (see screenshot) - take a look at [this](https://www.figma.com/file/1obAgfYZ639hpKXV1ZP7hJ/3.-FeatureSets-%3E-Case-intake?node-id=1801%3A208121) for reference
3. on the last page, the button text changes to 'Review case details' and the 'Clear page' button does not exist - can you look into adding support to enable this (i.e. allowing user to change the button text, optionally remove the tertiary button, pass in a different button function for the last page)?

bowenzhu1 (changes requested):
should remove the package-lock.json and yarn lock from the commit - will review more in detail in a bit

also, can we look more into/discuss how to save the entered information when navigating between steps? With the current implementation, the info gets lost when going to the next step, which is because we're no longer actually using the page specific submit functions

bowenzhu1:
closing in favor of #134

PR #102, opened 2022-11-09
Update Stepper Component To Allow Navigation

juliannejorda:
## Notion ticket link

[Update Stepper Component To Allow Navigation](https://www.notion.so/uwblueprintexecs/Update-stepper-component-to-allow-navigation-a75288035bc74878a914a81ac1a70f57)

## Implementation description
* Added the implementation where each step is clickable, and navigates to the correct page

## Steps to test
1. Run localhost:3000 and go to http://localhost:3000/intake
2. Test that you can click each stepper and it redirects to the correct intake page

* If it navigates to the correct page

## Checklist


frontend/src/components/intake/Stepper.tsx
```diff
@@ -29,28 +34,35 @@ const Stepper = ({ pages, activePage }: StepperProps): React.ReactElement => {
...
             <React.Fragment key={index}>
-              {isFirst || (
-                <>
-                  {/* connector */}
-                  <GridItem
-                    bg="gray.100"
-                    height="4px"
-                    alignSelf="center"
-                    gridRowStart={1}
+              <div
```
jennchenn, line 37: can we use `Box` instead

frontend/src/components/intake/Stepper.tsx
```diff
@@ -5,9 +5,14 @@ import { CheckCircle, Circle } from "react-feather";
...
   pages: string[];
   activePage: number;
+  setTheStep: (index: number) => void;
 };
 
-const Stepper = ({ pages, activePage }: StepperProps): React.ReactElement => {
+const Stepper = ({
+  pages,
+  activePage,
+  setTheStep,
```
jennchenn, line 14: nit: rename to `setStep`

jennchenn (changes requested):
navigation seems to work! i will confirm w designers in slack but i think:
1. we should make the cursor a pointer if the navigation buttons are clickable
2. maybe these should be implemented as buttons instead (so we get things like tabbing functionality)
3. we are currently able to navigate to any future step - i'm not sure if there is a way to prevent this if we don't currently store error state. thoughts?

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -35,6 +35,8 @@ const Intake = (): React.ReactElement => {
 
   const prevStep = () => setStep(step - 1);
 
+  const setTheStep = (index: number) => setStep(index + 1);
```
jennchenn, line 38: nit

```suggestion
  const setStep = (index: number) => setStep(index + 1);
```

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -105,6 +107,7 @@ const Intake = (): React.ReactElement => {
               "Individual details",
               "Program details",
             ]}
+            setStep={setTheStep}
```
jennchenn, line 110: nit

```suggestion
            setStep={setStep}
```

frontend/src/components/intake/Stepper.tsx
```diff
@@ -33,37 +33,36 @@ const Stepper = ({
...
-                    />
-                    <GridItem gridRowStart={2} />
-                  </>
-                )}
-                <GridItem color={color} gridRowStart={1} justifySelf="center">
-                  <Icon
-                    as={isComplete ? CheckCircle : Circle}
-                    width="24px"
-                    height="24px"
+            <Button
```
jennchenn, line 36: ahh i know i told you to do this but do you mind changing this to a `Box` with the `as='button'` prop? putting divs in buttons seems to be bad practice from a quick search

https://chakra-ui.com/docs/components/button#custom-button

jennchenn (changes requested):
so sorry changed my mind again about the button - hopefully a quick change but let me know if it isn't and i can create another ticket to just get this PR merged in!

jennchenn (approved):
nice!

PR #104, opened 2022-11-11
Create existing provider modal

jennchenn:
## Notion ticket link

[Create existing provider modal](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* created a new common modal component for use for future modals
* the isOpen and onClose parameters are passed in from a useDisclosure hook in the parent calling the modal component so that the parent can register when the modal is called based on a button click
* currently an example is in the homepage for the existing provider modal

## Steps to test
1. go to branch
2. go to localhost:3000/ and scroll all the way down
3. at the bottom click the open modal button to see the modal for the existing providers

* if the existing provider modal should be imported somewhere else (i couldn't find the provider screen)
* if the modal looks as expected

Screen Shot 2022-11-11 at 5 51 35 PM

## Checklist


frontend/src/components/intake/ExistingProviderModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+              <Spacer />
+              <Button
+                size="sm"
+                variant="tertiary"
+                leftIcon={<Icon as={UserPlus} />}
+              >
+                Add new provider
+              </Button>
+            </Flex>
+            <Select placeholder="Select provider">
```
kabiroberai, line 43: Just flagging this for your reference rn, but consider replacing this `Select` with a `CustomSelectField` once #88 is in.

jennchenn: @kabiroberai does the `CustomSelectField` support multi-select?

frontend/src/components/common/ModalComponent.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <Text paddingBottom="62px" textStyle="header-medium">
+          {primaryTitle}
+        </Text>
+        <ModalCloseButton marginTop="40px" marginRight="20px" size="sm" />
+        {modalContent}
+        <ModalFooter paddingTop="68px">
+          <Button variant="tertiary" onClick={onClose}>
+            Cancel
+          </Button>
+          <Button variant="primary" disabled onClick={onClick}>
```
jennchenn, line 51: once a user selects something, the button should no longer be disabled - i think we would need a way to determine whether the button should be disabled/not based on the contents

jennchenn (commented):
looks good!!

two things:
1. we need to be able to control whether primary button is disabled/not
2. the select component should be a multi-select--this can be left for another ticket though

jennchenn (commented):
there seems to be a bug with the disabled state: on first click the button is disabled on initial open; however if the user enters something, presses cancel, and reopens the modal, the button is still enabled--do you mind seeing if you can replicate this?

child provider

jennchenn (commented):
disabled button bug is fixed - i'm unable to approve the PR bc i opened it but it looks good to me

juthikahoque (approved):

PR #105, opened 2022-11-12
[kai/differentiate-child-behavior-familial-concerns] Change child_behaviors and familial_concerns

kai-wei-mo:
## Notion ticket link

[Differentiate child_behavior and familial_concerns](https://www.notion.so/uwblueprintexecs/Differentiate-child_behavior-and-familial_concerns-ee4487d1af224632a23ef1455fcdc277)

## Implementation description
* Exactly as the ticket describes

## Steps to test
1. Visual inspection

* N/A

## Checklist


backend/python/tests/functional/test_child_service.py
```diff
@@ -12,11 +12,6 @@
 from app.resources.child_dto import ChildDTO, CreateChildDTO
 from app.services.implementations.child_service import ChildService
 
-DUMMY_DAYTIME_CONTACT_DATA = {
-    "name": "Hamzaa Yusuff",
-    "contact_information": "8790832",
-    "dismissal_time": "4:00PM",
-}
```
kai-wei-mo, line 19: unrelated to pr. noticed that this var was not being used

backend/python/app/models/child.py
```diff
@@ -1,6 +1,13 @@
 from . import db
 from .base_mixin import BaseMixin
 
+child_join_child_behavior = db.Table(
```
jennchenn, line 4: nit

```suggestion
children_child_behaviors = db.Table(
```

backend/python/app/services/implementations/intake_service.py
```diff
@@ -8,3 +12,69 @@ def __init__(self, logger):
 
     def create_intake(self, intake):
         pass
+
+
+class FamilialConcernService(IFamilialConcernService):
```
jennchenn, line 17: hmm i know this is implemented as is due to the way i setup the ticket, but what is your take on keeping these services in `intake_service.py` and `child_service.py` vs. separating them out into new service files?

jennchenn (commented):
minor nits but i think this looks good!

backend/python/app/services/implementations/intake_service.py
kai-wei-mo: glad you asked! personally, i think having i would prefer having each service in their own file because it would be easier to navigate. for that matter, i would also prefer having each table be in its own file (e.g. `child` table is in `child.py`, and `children_child_behaviors` is in its own file) because if i have a table that handles the relationship between the `Foo` and `Bar` tables, it is unclear if this table should be declared in `foo.py` or `bar.py`

jennchenn: agree with this - let's move these into separate services + model files!

backend/python/app/models/__init__.py
```diff
@@ -6,19 +6,22 @@
...
+        daytime_contact,
+        familial_concern,
+        goal,
+        intake,
+        other_permitted_individual,
+        provider,
+        transportation_method,
+        user,
+        visit_location,
+    )
```
jennchenn, line 24: curious, why did you make this change?

backend/python/app/models/children_child_behaviors.py
```diff
@@ -0,0 +1,8 @@
+from . import db
+
+children_child_behaviors = db.Table(
+    "child_join_child_behavior",
```
jennchenn, line 4: nit i want to follow a similar convention for these tables without the join

e.g. `children_child_behaviors`--lmk if that's too confusing to you

jennchenn (commented):
just a few questions otherwise looks good!

backend/python/app/models/children_child_behaviors.py
kai-wei-mo: sure

backend/python/app/models/__init__.py
kai-wei-mo: oh, currently if one makes changes to an enum without touching any schema, alembic will not detect it. i thought my change would address this case but in fact it does not so i will revert it

PR #108, opened 2022-11-16
[kai/intake-post-route] Implement /intake routes and intake related services

kai-wei-mo:
## Notion ticket link

[Implement POST /intake route](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

## Implementation description
* Using the json input, create the necessary dependencies that an `intake` needs (e.g. `caregiver`s, `child`s, etc.) then create the `intake`
* If any dependency fails to be created (e.g. invalid fields) then roll back everything

* I am not 100% sure which fields in the JSON correspond to which models and which models' fields so I did some guessing as to what maps to what. Also, some mandatory fields I'm not sure about where they are in the JSON so I hardcoded stuff (e.g. email="test@mail.com") and also left some question mark comments in `routes/intake.py`

## Steps to test
Before continuing, please run:
```bash
docker-compose down --volumes
docker-compose up -d
docker exec -it cas_py_backend /bin/bash -c "flask db upgrade"
docker exec -it cas_py_backend /bin/bash -c "python -m tools.db_seed"
```

1. When POSTing with a correct payload, the intake and all relevant foreign dependencies are created (check that they exist in the db). One possible way to test is running `good.sh` (but this only checks the `intakes` db)
2. When POSTing with an incorrect payload (e.g. a mandatory field was left blank), not a single change is made to any table (check this as well). One possible way to test is running `bad.sh` (but this only checks the `intakes` db)

* 

## Checklist


backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+    except Exception as error:
+        return jsonify(error), 400
 
 
+# create an intake
 @blueprint.route("/", methods=["POST"], strict_slashes=False)
-@require_authorization_by_role({"User", "Admin"})
-@validate_request("IntakeDTO")
+# @require_authorization_by_role({"User", "Admin"})
+# @validate_request("CreateIntakeDTO")
```
jennchenn, line 52: does the validation not work for this?

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+            service, fn, arg = undo
+            service.__dict__[fn](arg)
+
+    # intake_id
+    intake_response = intake_service.get_all_intakes()
+    intake_id = len(intake_response) + 1
+
+    # intake
+    intake = {
+        "user_id": 1,  # ?
```
jennchenn, line 67: this should come from the request data

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+        "intake_status": "ACCEPTED",
```
jennchenn, line 68: default status should be `IN_PROGRESS`

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+        "scheduling_requirements": request.json["programDetails"][
+            "schedulingRequirements"
+        ],
+        "suggested_start_date": request.json["programDetails"]["suggestedStartDate"],
+        "date_accepted": "2020-01-01",  # ?
+        "access_weekday": None,  # ?
+        "access_location": None,  # ?
+        "access_time": None,  # ?
+        "lead_access_worker_id": 1,  # ?
+        "denial_reason": None,  # ?
```
jennchenn, line 95: these fields should be updated later - is there an issue with leaving them blank?

backend/python/app/rest/bad.sh
```diff
@@ -0,0 +1 @@
+curl -X POST -H "Content-Type: application/json" -d @bad.json http://localhost:5000/intake > bad.html
```
jennchenn, line 1: are these scripts going to be removed prior to merging / do they get run somewhere in the test suite?

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+        except Exception as error:
+            run_undos()
+            return jsonify(error), 400
+
+    # familial concerns
+    familialConcerns = request.json["programDetails"]["familialConcerns"]
+    for familialConcern in familialConcerns:
+        familialConcern = {
+            "concern": familialConcern,
+            "is_default": False,  # ?
```
jennchenn, line 162: `False` is correct

perhaps the service https://github.com/uwblueprint/childrens-aid-society/blob/540c85cc1670ea406890dd6de55f3ea821a273e8/backend/python/app/services/implementations/familial_concern_service.py#L48-L59 should make a call to `get_familial_concern` first to verify it doesn't add the same concern in twice

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+
+    # short term goals
+    shortTermGoals = request.json["programDetails"]["shortTermGoals"]
+    for shortTermGoal in shortTermGoals:
+        shortTermGoal = {
+            "type": "SHORT_TERM",
+            "goal": shortTermGoal,
+            # "start_date": None,
+            # "end_date": None,
+            # "is_default": False,
```
jennchenn, line 189: i think these commented lines can be removed if those are the default values already

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+        except Exception as error:
+            run_undos()
+            return jsonify(error), 400
+
+    # goals
+
+    # short term goals
+    shortTermGoals = request.json["programDetails"]["shortTermGoals"]
+    for shortTermGoal in shortTermGoals:
+        shortTermGoal = {
```
jennchenn, line 184: i would suggest naming this something else to avoid confusion (since you use `for shortTermGoal in shortTermGoals` already

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+
+        # children
+        child_obj = {
+            "intake_id": intake_id,
+            "first_name": child["childInfo"]["name"].rsplit(" ", 1)[0],
+            "last_name": child["childInfo"]["name"].rsplit(" ", 1)[1],
+            "date_of_birth": child["childInfo"]["dateOfBirth"],
+            "cpin_number": child["childInfo"]["cpinFileNumber"],
+            "child_service_worker_id": 1,  # ?
+            "daytime_contact_id": 1,  # ?
```
jennchenn, line 248: `daytime_contact_id` comes from the above daytimeContact that was added

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+                    "delete_daytime_contact",
+                    daytimeContact_response.id,
+                )
+            )
+        except Exception as error:
+            run_undos()
+            return jsonify(error), 400
+
+        # children
+        child_obj = {
```
jennchenn, line 241: my bad some updates need to be made to child model
1. combine `first_name` and `last_name` to just `name`
2. remove `child_service_worker_id` and replace with `serviceWorker`
3. remove `has_kinship_provider` and `has_foster_placement` (i think these can be inferred from the rest of the data)

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+        # provider
+        providers = child["provider"]
+        for provider in providers:
+            provider_obj = {
+                "name": provider["name"],
+                "file_number": provider["fileNumber"],
+                "primary_phone_number": provider["primaryPhoneNumber"],
+                "secondary_phone_number": provider["secondaryPhoneNumber"],
+                "email": "test@mail.com",
+                "address": "1234 Test St",
```
jennchenn, line 269: these should come from the data as well

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+                )
+                undos.append(
+                    (provider_service, "delete_provider", provider_response.id)
+                )
+            except Exception as error:
+                run_undos()
+                return jsonify(error), 400
+
+        # concerns
+        concerns = child["childInfo"]["concerns"]
```
jennchenn, line 286: this should map to `ChildBehavior`

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+                    "delete_other_permitted_individual",
+                    permittedIndividual_response.id,
+                )
+            )
+        except Exception as error:
+            run_undos()
+            return jsonify(error), 400
+
+    # familial concerns
+    familialConcerns = request.json["programDetails"]["familialConcerns"]
```
jennchenn, line 158: an entry would also need to be added to `intakes_concerns`

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+            )
+        except Exception as error:
+            run_undos()
+            return jsonify(error), 400
+
+    # goals
+
+    # short term goals
+    shortTermGoals = request.json["programDetails"]["shortTermGoals"]
+    for shortTermGoal in shortTermGoals:
```
jennchenn, line 183: an entry would also need be to added to `intakes_goals`

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+        }
+        try:
+            shortTermGoal_response = goal_service.add_new_goal(**shortTermGoal)
+            undos.append((goal_service, "delete_goal", shortTermGoal_response.id))
+        except Exception as error:
+            run_undos()
+            return jsonify(error), 400
+
+    # long term goals
+    longTermGoals = request.json["programDetails"]["longTermGoals"]
```
jennchenn, line 199: same comments as for shortTermGoals

backend/python/app/rest/intake_routes.py
```diff
@@ -1,16 +1,291 @@
...
+        return jsonify(error), 400
 
 
+# create an intake
 @blueprint.route("/", methods=["POST"], strict_slashes=False)
-@require_authorization_by_role({"User", "Admin"})
-@validate_request("IntakeDTO")
+# @require_authorization_by_role({"User", "Admin"})
+# @validate_request("CreateIntakeDTO")
 def create_intake():
```
jennchenn, line 53: not sure if it'd be better to move all this logic out into a service rather than polluting this file so it's easier to see all the intake routes at a glance here - thoughts?

jennchenn (changes requested):
lots of good work here!! did a pass through the `intake_routes.py` file to hopefully clear up some data issues

let me know what seems reasonable to tackle in this PR

bowenzhu1:
i should probably clean up the migrations

PR #123, opened 2023-01-22
Implement ‘Providers’ box within Add Child

Ezzhingy:
## Notion ticket link

[Implement ‘Providers’ box within Add Child] (https://www.notion.so/uwblueprintexecs/134e1f159caa4f73939d11a700c325ad?v=39cf7d5a98384eb9af2bff2265d0ac8a&p=fbcf786463c244adbcaf045b7ed8a41e&pm=s)

## Implementation description
* Create DoublePromptBox to house two buttons instead of one
Screenshot 2023-01-22 at 5 49 41 PM

## Steps to test
1. Go to http://localhost:3000/intake/add-child and check to make sure providers box is working

* providers box implementation

## Checklist

### DIFF
--- a/frontend/src/components/intake/PromptBox.tsx
+++ b/frontend/src/components/intake/PromptBox.tsx
@@ -51,14 +51,16 @@ const PromptBox = ({
         >
           {buttonText}
         </Button>
-        <Button
-          variant="secondary"
-          colorScheme="gray.600"
-          leftIcon={secondaryButtonIcon}
-          onClick={secondaryOnButtonClick}
-        >
-          {secondaryButtonText}
-        </Button>
+        {secondaryButtonText && (
+          <Button
+            variant="secondary"
+            colorScheme="gray.600"
+            leftIcon={secondaryButtonIcon}
+            onClick={secondaryOnButtonClick}
+          >
+            {secondaryButtonText}
+          </Button>
+        )}
       </HStack>
     </VStack>
   );
--- a/frontend/src/components/intake/PromptBox.tsx
+++ b/frontend/src/components/intake/PromptBox.tsx
@@ -5,8 +5,11 @@ export type PromptBoxProps = {
   headerText: string;
   descriptionText: string;
   buttonText: string;
-  buttonIcon: ReactElement;
+  buttonIcon?: ReactElement;
   onButtonClick: () => void;
+  secondaryButtonText?: string;
+  secondaryButtonIcon?: ReactElement;
+  secondaryOnButtonClick?: () => void;
 };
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/intake/child-information/ChildProviderForm.tsx
```diff
@@ -15,18 +17,30 @@ const ChildProviderForm = (): React.ReactElement => {
...
+    <>
+      <VStack padding="100px">
+        <DoublePromptBox
+          headerText="Providers"
+          descriptionText="At least one provider must be indicated for each child"
+          buttonText1="Add new provider"
+          buttonIcon1={<Icon as={UserPlus} w="16px" h="16px" />}
+          onButtonClick1={onOpenNewProviders}
+          buttonText2="Select providers"
+          buttonIcon2={<Icon as={UserPlus} w="0px" margin="0px" />}
```
bowenzhu1, line 29: I think we should just make this field optional (edit `DoublePromptBoxPrompts` in the other file to have `buttonIcon2?: string`) and then not include it in this call here instead of passing a 0px icon

frontend/src/components/intake/PromptBox.tsx
```diff
@@ -48,4 +48,67 @@ const PromptBox = ({
   );
 };
 
+export type DoublePromptBoxProps = {
+  headerText: string;
+  descriptionText: string;
+  buttonText1: string;
```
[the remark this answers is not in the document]
bowenzhu1: same for the buttonIcon, and onButtonClick

bowenzhu1 (changes requested):
good job! One other major change I would make is to either separate your DoublePromptBox component into its own file (same folder). Or to try and just combine the logic with the existing PromptBox by making some props optional and conditionally render some components like the secondary button

frontend/src/components/intake/PromptBox.tsx
```diff
@@ -36,14 +42,24 @@ const PromptBox = ({
...
+      <HStack>
+        <Button
+          variant="tertiary"
+          colorScheme="gray.600"
+          leftIcon={buttonIcon}
+          onClick={onButtonClick}
+        >
+          {buttonText}
+        </Button>
+        <Button
```
bowenzhu1, line 54: we should conditionally render the second button, maybe based on `{secondaryButtonText &&`. Otherwise single button PromptBoxes just have an empty second button (see the Individual Details - Add Child box).

bowenzhu1 (changes requested):
one more small change to make. The code is def cleaner now tho after combining the components - good stuff

PR #124, opened 2023-01-23
Add footer with save button to child form

IvanYu327:
## Notion ticket link

[Add ‘Save Child Information’ button to Add Child flow](https://www.notion.so/uwblueprintexecs/Add-Save-Child-Information-button-to-Add-Child-flow-f6130403d9fe4e71ac2e490ba2ce40f1)

## Implementation description
- Added footer to `AddChildPage.tsx` with save button since it makes more sense to have on save button instead of three (one on each form page). Also the Figma said "Primary button is disabled until all required fields across the three pages have valid inputs. "
- Added onKeyUp listeners to the Fields on the first form that update the state. This is useful for the future if format validation is implemented.
- Added check for required fields for the child Information. The other two forms are incomplete so did not add checks for that but left comments.

## Steps to test
1. Open the form `http://localhost:3000/intake/add-child`
2. Fill in the form
3. After filling in name, cpin, and DOB, the save information button should become clickable
4. Deleting any of the above fields disables it

*  `AddChildPage.tsx` -> footer
*   `ChildInformationForm.tsx` -> added onKeyUp listeners 

## Checklist

### DIFF
--- a/frontend/src/components/intake/child-information/AddChildPage.tsx
+++ b/frontend/src/components/intake/child-information/AddChildPage.tsx
@@ -21,7 +21,7 @@ const AddChild = (): React.ReactElement => {
     childBehaviours: "",
   });
 
-  const requiredInfomationMissing : boolean = 
+  const requiredInfomationMissing : boolean =
     !childDetails.childName || 
     !childDetails.cpinFileNumber ||
     !childDetails.dateOfBirth 
@@ -82,7 +82,12 @@ const AddChild = (): React.ReactElement => {
       </VStack>
       {renderChildForm()}
 
-      <Box bg="white" height="87px" display="flex" justifyContent="flex-end" alignItems="center"
+      <Box 
+        bg="white"
+        height="87px"
+        display="flex"
+        justifyContent="flex-end"
+        alignItems="center"
         shadow="0px -4px 12px rgba(226, 225, 236, 0.4), 0px -8px 24px rgba(226, 225, 236, 0.25)"
       >
         <Button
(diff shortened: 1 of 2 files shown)


frontend/src/components/intake/child-information/AddChildPage.tsx
```diff
@@ -69,7 +81,24 @@ const AddChild = (): React.ReactElement => {
         />
       </VStack>
       {renderChildForm()}
-      {/* TODO: Add footer component */}
+
+      <Box
```
bowenzhu1, line 85: I think you need to change the CSS here/ in the parent component, so that this footer sticks to the bottom. If you zoom out and make the screen smaller, or in School/Daycare info, it floats.

frontend/src/components/intake/child-information/AddChildPage.tsx
```diff
@@ -69,7 +81,24 @@ const AddChild = (): React.ReactElement => {
...
+        bg="white"
+        height="87px"
+        display="flex"
+        justifyContent="flex-end"
+        alignItems="center"
+        shadow="0px -4px 12px rgba(226, 225, 236, 0.4), 0px -8px 24px rgba(226, 225, 236, 0.25)"
+      >
```
bowenzhu1, line 92: can we also add a // TODO here for a cancel button

bowenzhu1 (changes requested):
looks good, one pretty minor fix to make regarding CSS for the footer.

I like the implementation , but just wondering if there's a specific reason you had in mind for using onKeyUp vs onChange or onInput. Either one should work, but just curious.

Example of footer floating:

image

frontend/src/components/intake/child-information/AddChildPage.tsx
bowenzhu1: looks like the footer blocks off some of the page now (screenshot below). Maybe add margin-bottom/change the css for the page components?

bowenzhu1 (changes requested):
looks like the footer blocks off some of the page now. Maybe add margin bottom/modify the css for the page components themselves?

image

bowenzhu1 (approved):
lgtm!

PR #134, opened 2023-02-04
Intake footer

TW1234567:
## Notion ticket link

[Create intake form footer navigation component](https://www.notion.so/uwblueprintexecs/134e1f159caa4f73939d11a700c325ad?v=39cf7d5a98384eb9af2bff2265d0ac8a&p=3e8a3cb2f18742fa91e7f69326a78bdc&pm=s)

## Implementation description
* Create footer component which is places in each individual registration page. 
* Footer handles navigation and saving state on click
* Add header stepper to each individual page
* Stepper handle navigation and saving state on click

## Steps to test
1. Edit fields and navigate via either stepper or footer. Notice that the fields (except for Individual details) page is being updated.
[Untitled_ Feb 4, 2023 3_44 AM.webm](https://user-images.githubusercontent.com/117593511/216758018-2a22fa2a-dd64-4031-9d58-b29642c89d5d.webm)

* I didn't handle saving state for individual details because I felt the implementation was rather involved and out of the scope of this PR 
* I noticed that Eugene had worked on some button features for hiding buttons on read only. I remove those code bits and just added buttons via the footer
* The need to have a previous button within the footer feels rather important to me. Especially int he case of the the review page where a header stepper is not provided, I feel that the UI would look clearer having the previous button in the footer and not in the nested page component. 
* I didn't implement the clear page button feature. This was because the previous [PR being built on-top of](https://github.com/uwblueprint/childrens-aid-society/pull/95) didn't mention such requirement.
* Since I had to put the header stepper inside of the registration page, the stepper inherited some additional css properties which reduced the width to the width of the page element. I couldn't find a clean way to increase the width back to normal.

## Checklist


frontend/src/components/intake/IntakeFormFooter.tsx
```diff
@@ -0,0 +1,90 @@
...
+        height="48px"
+        loadingText="Submitting"
+        type="submit"
+        isLoading={registrationLoading}
+        onClick={() => {
+          onNextStep();
+        }}
+      >
+        {nextBtnTxt}
+      </Button>
```
bowenzhu1, line 85: can we add the right arrow icon beside the Next section button? ex. https://www.figma.com/file/1obAgfYZ639hpKXV1ZP7hJ/3.-FeatureSets-%3E-Case-intake?node-id=1412%3A779304&t=Wf9212d8RydGMplg-0

frontend/src/components/intake/ReviewCaseForm.tsx
```diff
@@ -54,6 +63,15 @@ const ReviewForm = ({
           Previous Button
         </Button>
       </Box>
+
+      <IntakeFooter
+        currentStep={IntakeSteps.REVIEW_CASE_DETAILS}
+        nextBtnTxt="Next"
```
bowenzhu1, line 69: this should be Submit case, and we should hide the clear button.

I don't think we **need** a previous button, since there are edit buttons to direct the client to previous steps.

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -48,6 +47,7 @@ const Intake = (): React.ReactElement => {
             referralDetails={referralDetails}
             setReferralDetails={setReferralDetails}
             nextStep={nextStep}
+            setStep={setStep}
           />
```
bowenzhu1, line 51: hm, for this entire function and files in general, can we make wrapper components?

Ex. instead of calling ReferralForm we call ReferralFormStep, where ReferralFormStep is basically just:
`<Stepper>
        <ReferralForm>
</Stepper>`

ps. also see the comment below for the Container/stepper sizing

This will make it so we can reuse the ReferralForm components elsewhere. like in the Review screen. Currently it has the stepper in it lol 
image

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -107,20 +106,6 @@ const Intake = (): React.ReactElement => {
...
-              "Individual details",
-              "Program details",
-            ]}
-            setStep={setStep}
-            activePage={step}
-          />
-        ) : (
-          <></>
-        )}
         <Container maxWidth="container.xl" padding="30px 96px">
```
bowenzhu1, line 109: if we change the renderDetailsForm to call wrapper components (mentioned in above comment), we can maintain the original stepper size by removing this container here, and just calling it in each wrapper component I think?

So now the wrapper ReferralFormStep could look like 
`<Stepper><Container> <ReferralForm>...  </Container> </Stepper>`

frontend/src/components/intake/CourtInformationForm.tsx
```diff
@@ -53,121 +56,133 @@ const CourtInformationForm = ({
...
+                name="firstNationBand"
+              />
+            </FormControl>
+          </HStack>
+        </Form>
+      </FormikProvider>
+      <IntakeFooter
+        currentStep={IntakeSteps.COURT_INFORMATION}
+        nextBtnTxt="Next"
+        showClearPageBtn={!!true}
```
bowenzhu1, line 180: is this boolean cast !! needed? I think just passing {true} should work

frontend/src/components/intake/CourtInformationForm.tsx
```diff
@@ -53,121 +56,133 @@ const CourtInformationForm = ({
...
+        isStepComplete={() => true}
```
bowenzhu1, line 181: lol this makes sense but just looks funny rn. can we add a comment here along the lines of // TODO: validate form

frontend/src/components/intake/ReviewCaseForm.tsx
```diff
@@ -3,20 +3,28 @@ import { Box, Button, HStack, Text, Icon, Stack } from "@chakra-ui/react";
...
 
 const ReviewForm = ({
   referralDetails,
   setReferralDetails,
   nextStep,
   prevStep,
+  setStep,
 }: ReviewFormProps): React.ReactElement => {
+  const onNextStep = () => {
+    nextStep();
```
bowenzhu1, line 25: let's add a TODO here, since this nextStep() doesn't do anything right now

bowenzhu1 (changes requested):
phew thanks for putting this together! looks good overall - i addressed most of your concerns + requested some changes (such as adding wrappers). lmk if you have questions

oh also sorry for the merge conflict lol, I just edited the ReferralForm css grid spacing a little in my merged PR

frontend/src/components/intake/CourtInformationForm.tsx
TW1234567: Nope, will be removed

frontend/src/components/intake/ReferralForm.tsx
```diff
@@ -27,106 +29,140 @@ type ReferralFormProps = {
...
-          </Button>
-        )}
-      </Form>
-    </Formik>
+          </FormControl>
+        </Form>
+      </FormikProvider>
+      <IntakeFooter
+        currentStep={IntakeSteps.CASE_REFERAL}
+        nextBtnTxt="Next"
```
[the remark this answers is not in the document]
TW1234567: A little confused. I think "nextBtnTxt" should be "Next"

frontend/src/components/intake/IntakeFormFooter.tsx
```diff
@@ -0,0 +1,90 @@
...
+      align="center"
+      justify={{ sm: "center", md: "end" }}
+      flexWrap="wrap"
+      padding="20px"
+      left="0"
+      position="fixed"
+      bottom="0"
+      zIndex="5"
+    >
+      {showClearPageBtn ? (
```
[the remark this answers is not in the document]
TW1234567: I'll change to `showClearPageBtn && <Button>`

bowenzhu1 (approved):
looks good!! maybe we'll revisit the stepper width thing, but this is good to merge. good work and thanks for debugging thru everything

PR #136, opened 2023-02-06
Review Case - Court Information

xiongjasmine:
## Notion ticket link
[Review Case - Court Information](https://www.notion.so/uwblueprintexecs/Review-Case-Court-Information-94c0781aa66f4520a2da35df539168e2)

## Implementation description
* Implemented UI for Court Information on the Case Review page
* Modified other components like Autocomplete and CustomSelectField to include optional readOnly parameter
Screen Shot 2023-02-06 at 11 06 58 AM

## Steps to test
1. Go to localhost:3000/intake, fill out the form (especially the Court Information section), go to review
2. Check if design is correct
3. Make sure user can modify and interact with the fields on the Court Information page
4. Make sure user cannot modify or interact with the fields on the Case Review page

* Design
* Check the modifications for Autocomplete.tsx and CustomSelectField.tsx
* Not sure if the conditional cases (eg. 'Attach document' vs 'Download attachment' implementation in CourtInformationForm.tsx) could be handled more elegantly, thoughts on this syntax?

## Checklist

### DIFF
--- a/frontend/src/components/intake/Autocomplete.tsx
+++ b/frontend/src/components/intake/Autocomplete.tsx
@@ -16,6 +16,7 @@ import CustomTag from "../common/CustomTag";
 export type AutocompleteFieldProps = CustomInputProps & {
   name: string;
   hints: string[];
+  readOnly?: boolean;
 };
 
 export const AutocompleteSuggestions = ({
@@ -59,6 +60,7 @@ export const AutocompleteSuggestions = ({
 export const AutocompleteField = ({
   name,
   hints,
+  readOnly = false,
   ...props
 }: AutocompleteFieldProps): React.ReactElement => {
   const [field, , helpers] = useField<string>(name);
@@ -79,13 +81,16 @@ export const AutocompleteField = ({
             type="string"
             onFocus={() => setFocus(true)}
             onBlur={() => setFocus(false)}
+            isReadOnly={readOnly}
           />
         </FormControl>
       </PopoverTrigger>
-      <AutocompleteSuggestions
-        hints={filtered}
-        onSelect={(value) => helpers.setValue(value)}
-      />
+      {!readOnly && (
+        <AutocompleteSuggestions
+          hints={filtered}
+          onSelect={(value) => helpers.setValue(value)}
+        />
+      )}
     </Popover>
   );
 };
--- a/frontend/src/components/intake/CourtInformationForm.tsx
+++ b/frontend/src/components/intake/CourtInformationForm.tsx
@@ -9,7 +9,7 @@ import {
   Icon,
 } from "@chakra-ui/react";
 import { Field, Form, Formik } from "formik";
-import { FilePlus, ChevronDown } from "react-feather";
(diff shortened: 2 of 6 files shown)


frontend/src/components/intake/CourtInformationForm.tsx
```diff
@@ -106,13 +110,20 @@ const CourtInformationForm = ({
...
                   style={{ cursor: "pointer" }}
                 />
-                <Button
-                  onClick={handleClick}
-                  variant="tertiary"
-                  leftIcon={<FilePlus />}
-                >
-                  Attach document
-                </Button>
+                {!readOnly && (
```
bowenzhu1, line 113: we could use a ternary operator here `{readOnly ? <Button/> : <Button/>}` instead of two && checks

frontend/src/components/intake/Autocomplete.tsx
```diff
@@ -16,6 +16,7 @@ import CustomTag from "../common/CustomTag";
 export type AutocompleteFieldProps = CustomInputProps & {
   name: string;
   hints: string[];
+  readOnly?: true;
```
bowenzhu1, line 19: We should use `boolean` as the type here instead of true, I might have missed this in Eugene's PR

frontend/src/components/intake/Autocomplete.tsx
```diff
@@ -59,6 +60,7 @@ export const AutocompleteSuggestions = ({
 export const AutocompleteField = ({
   name,
   hints,
+  readOnly,
```
bowenzhu1, line 63: we can pass in a default value of false `readOnly = false` here since that should be the default behaviour

frontend/src/components/intake/CourtInformationForm.tsx
```diff
@@ -27,13 +27,15 @@ type CourtInformationFormProps = {
   setCourtDetails: React.Dispatch<React.SetStateAction<CourtDetails>>;
   nextStep: () => void;
   prevStep: () => void;
+  readOnly?: true;
```
bowenzhu1, line 30: same as above, should use `boolean`, and also add default value of false in props below

frontend/src/components/intake/CustomSelectField.tsx
```diff
@@ -20,6 +20,7 @@ export type CustomSelectProps = CustomInputProps & {
   icon?: JSX.Element;
   iconRight?: JSX.Element;
   options: string[];
+  readOnly?: true;
```
bowenzhu1, line 23: I think Deen also changed this file, so might run into merge conflicts.

We can change this to the same as the above comments too though (boolean, default value false)

bowenzhu1 (changes requested):
some small/style changes, but looks good overall!

there's some merge conflicts because Deen worked on the Access Program equivalent of this ticket. I think they should be pretty simple to resolve by just accepting both incoming changes?

bowenzhu1 (approved):
good work!

PR #145, opened 2023-02-19
Implement “View and edit details” button for Providers and Caregivers boxes

Ezzhingy:
## Notion ticket link

[Implement “View and edit details” button for Providers and Caregivers boxes](https://www.notion.so/uwblueprintexecs/Implement-View-and-edit-details-button-for-Providers-and-Caregivers-boxes-63fb7f9ed5364ea2b1ebf5cd71609719)

## Implementation description
* Implement functionality for "View and Edit Details" btn for Visiting Family Members
* When btn is clicked, opens up existing modal with info pre-filled
* User can then edit and click "Save visiting family member" to change that specific entry

## Steps to test
1. Go to `/intake` --> Individual details
2. Mess around with adding families, editing families, editing different families, etc.

* Code, my implementation is messy, looking for a more elegant solution

## Checklist

### DIFF
--- a/frontend/src/components/intake/IndividualDetails.tsx
+++ b/frontend/src/components/intake/IndividualDetails.tsx
@@ -22,7 +22,7 @@ const IndividualDetails = ({
           padding={1}
           borderRadius="5px"
         >
-          Children
+          CHILDREN
         </Text>
         {childrenDetails.map((indiv, i) => (
           <VStack key={i} w="full">
--- a/frontend/src/components/intake/IndividualDetails.tsx
+++ b/frontend/src/components/intake/IndividualDetails.tsx
@@ -1,19 +1,7 @@
 import React from "react";
-import {
-  Button,
-  VStack,
-  Text,
-  HStack,
-  Icon,
-  Divider,
-  useDisclosure,
-} from "@chakra-ui/react";
+import { Button, VStack, Text, HStack, Icon, Divider } from "@chakra-ui/react";
 import { ArrowRight } from "react-feather";
 import { IndividualDetailsOverview } from "./PromptBox";
-import NewCaregiverModal, {
-  Caregivers,
-  CaregiverDetails,
-} from "./NewCaregiverModal";
 
 type IndividualDetailsProps = {
   childrenDetails: IndividualDetailsOverview[];
--- a/frontend/src/components/intake/IndividualDetails.tsx
+++ b/frontend/src/components/intake/IndividualDetails.tsx
@@ -1,8 +1,19 @@
 import React from "react";
-import { Button, VStack, Text, HStack, Icon, Divider } from "@chakra-ui/react";
+import {
+  Button,
+  VStack,
+  Text,
+  HStack,
+  Icon,
+  Divider,
+  useDisclosure,
+} from "@chakra-ui/react";
 import { ArrowRight } from "react-feather";
-import { useHistory } from "react-router-dom";
 import { IndividualDetailsOverview } from "./PromptBox";
+import NewCaregiverModal, {
+  Caregivers,
+  CaregiverDetails,
+} from "./NewCaregiverModal";
 
 type IndividualDetailsProps = {
   childrenDetails: IndividualDetailsOverview[];
@@ -13,8 +24,6 @@ const IndividualDetails = ({
   childrenDetails,
   caregiverDetails,
 }: IndividualDetailsProps): React.ReactElement => {
-  const history = useHistory();
-
   return (
     <>
       <VStack paddingY="32px" align="flex-start" spacing="16px">
@@ -43,8 +52,7 @@ const IndividualDetails = ({
                 textStyle="button-small"
                 variant="tertiary"
                 onClick={() => {
-                  history.goBack();
-                  // TODO: implement view and edit details button, history.goBack() is just a placeholder, replace when ready
+                  // TODO: implement view and edit details button
                 }}
                 rightIcon={<Icon as={ArrowRight} h="16px" />}
               >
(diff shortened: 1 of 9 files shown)


frontend/src/components/intake/NewCaregiverModal.tsx
```diff
@@ -50,6 +56,38 @@ const NewCaregiverModal = ({
     onClose();
   };
 
+  useEffect(() => {
```
bowenzhu1, line 59: instead of this use effect, we can try to just set the initial values when we define the useStates (line 38)

frontend/src/components/intake/NewCaregiverModal.tsx
```diff
@@ -22,12 +22,18 @@ type NewCaregiverProps = {
   isOpen: boolean;
   onClick: (newCaregiver: CaregiverDetails) => void;
   onClose: () => void;
+  caregivers: Caregivers;
```
bowenzhu1, line 25: would it be better to just pass in the specific caregiver instead of the entire array + indexValue?

bowenzhu1: and then for new/blank cases, we can define a default dummy caregiver with empty "" values for every field

frontend/src/components/intake/NewCaregiverModal.tsx
```diff
@@ -22,12 +22,18 @@ type NewCaregiverProps = {
...
+  indexValue: number;
+  setEdit: React.Dispatch<React.SetStateAction<boolean>>;
```
bowenzhu1, line 27: I don't think we need setEdit passed here

frontend/src/components/intake/NewCaregiverModal.tsx
bowenzhu1: see above comment about passing in a caregiver in the props, which we could use when defining initial state values

frontend/src/components/intake/indivDetails/CaregiverProviderForm.tsx
```diff
@@ -17,14 +17,21 @@ const CaregiverForm = ({
   setCaregivers,
 }: CaregiverFormProps): React.ReactElement => {
   const [caregiversDeleted, setCaregiversDeleted] = useState(0);
+  const [indexValue, setIndexValue] = useState(-1);
+  const [edit, setEdit] = useState(false);
```
bowenzhu1, line 21: this edit state seems redundant - seems like we can check everything w the indexValue instead

frontend/src/components/intake/indivDetails/CaregiverProviderForm.tsx
```diff
@@ -17,14 +17,21 @@ const CaregiverForm = ({
...
+
   const {
     onOpen: onOpenAddCaregivers,
     isOpen: isOpenAddCaregivers,
     onClose: onCloseAddCaregivers,
   } = useDisclosure();
 
   const onClickNewCaregiver = (newCaregiver: CaregiverDetails) => {
-    caregivers.push(newCaregiver);
+    if (edit) {
```
bowenzhu1, line 30: based on above comment could do:
`if (indexValue >= 0) {
caregivers.splice(indexValue, 1, newCaregiver);
  } else {
    caregivers.push(newCaregiver);
  }
`

frontend/src/components/intake/indivDetails/CaregiverProviderForm.tsx
```diff
@@ -17,14 +17,21 @@ const CaregiverForm = ({
   setCaregivers,
 }: CaregiverFormProps): React.ReactElement => {
   const [caregiversDeleted, setCaregiversDeleted] = useState(0);
+  const [indexValue, setIndexValue] = useState(-1);
```
bowenzhu1, line 20: let's rename this to something like `selectedIndex` for a bit more clarity

bowenzhu1 (changes requested):
the index passing approach works, although it's a bit messy since the modal component isn't in the same level as the promptbox (but it's fine)

i don't think we should have useEffect though. I left some comments to help clean up some of the logic. lmk if you have questions

apologies for the late review!

frontend/src/components/intake/IndividualDetails.tsx
```diff
@@ -25,7 +22,7 @@ const IndividualDetails = ({
           padding={1}
           borderRadius="5px"
         >
-          CHILDREN
+          Children
```
bowenzhu1, line 25: let's keep this in caps based on designs

frontend/src/components/intake/NewProviderModal.tsx
```diff
@@ -196,11 +244,15 @@ const NewProviderModal = ({
...
+              ? providerFileNo
+              : provider?.providerFileNo) &&
+            (primaryPhoneNoChanged
+              ? primaryPhoneNo
+              : provider?.primaryPhoneNo) &&
+            (addressChanged ? address : provider?.address) &&
+            (relationshipChanged ? relationship : provider?.relationship)
           )
         }
         secondaryTitle="Individual Details"
```
bowenzhu1, line 258: can we add a check to change the button title below to "Save provider" instead of "Add provider" if a provider is passed in

frontend/src/components/intake/child-information/ChildProviderForm.tsx
```diff
@@ -52,19 +58,31 @@ const ChildProviderForm = ({
...
           descriptionText="At least one provider must be indicated for each child"
-          buttonText="Select providers"
-          onButtonClick={onOpenExistingProviders}
-          secondaryButtonText="Add new provider"
+          buttonText="Add new provider"
+          onButtonClick={onOpenNewProviders}
+          secondaryButtonText="Select providers"
           secondaryButtonIcon={<Icon as={UserPlus} w="16px" h="16px" />}
-          secondaryOnButtonClick={onOpenNewProviders}
+          secondaryOnButtonClick={onOpenExistingProviders}
```
bowenzhu1, line 82: i think we want the way it was before? Add new provider should be the left button based on the designs

bowenzhu1 (changes requested):
some small changes relating to ui/copy, but good to go otherwise!

bowenzhu1 (approved):
lgtm! great work debugging and reaching out w questions

PR #152, opened 2023-02-27
implement 'Other Permitted Individuals' add and delete

xiongjasmine:
## Notion ticket link

[Integrate Other Permitted Individuals add and delete](https://www.notion.so/uwblueprintexecs/Integrate-Other-Permitted-Individuals-add-and-delete-5d1197d13fa74173ae5bec614d217df3)

## Implementation description
* Implemented add and delete for 'Other Permitted Individuals' section
* Added `DeleteConfirmationModal.tsx` which is a modal that asks the user if they're sure they wanna delete the item
* 'View and edit details' implementation is not completed - this is still WIP so I may do this is a separate PR

## Steps to test
1. Go to `/intake` and go to 'Program Details' in the stepper
2. Click the 'Add' button at the bottom
3. Enter individual info and save it
4. Try adding more individuals and try deleting them too

* Check that you are able to add individuals
* Make sure the correct UI appears when the list of permitted individuals is empty
* Check that the delete button makes a modal pop up and check that the name in the modal is correct
* Check that the right individual deletes

## Checklist

### DIFF
--- a/frontend/src/components/intake/DeleteConfirmationModal.tsx
+++ b/frontend/src/components/intake/DeleteConfirmationModal.tsx
@@ -1,0 +1,45 @@
+import React from "react";
+import { Box, Text } from "@chakra-ui/react";
+import ModalComponent from "../common/ModalComponent";
+
+type DeleteConfirmationProps = {
+  isOpen: boolean;
+  onClick: () => void;
+  onClose: () => void;
+  providerName: string;
+};
+
+const DeleteConfirmationModal = ({
+  isOpen,
+  onClick,
+  onClose,
+  providerName,
+}: DeleteConfirmationProps): React.ReactElement => {
+  return (
+    <Box>
+      <ModalComponent
+        primaryTitle="Remove permitted individual"
+        secondaryTitle="Individual Details"
+        modalContent={
+          <Text>
+            This will remove the individual {providerName} from this case. You
+            will need to re-enter their information to add them to the case
+            intake file.
+          </Text>
+        }
+        onClick={() => {
+          onClick();
+          onClose();
+        }}
+        isOpen={isOpen}
+        onClose={() => {
+          onClose();
+        }}
+        disabled={false}
+        primaryButtonTitle="Remove permitted individual"
+      />
+    </Box>
+  );
+};
+
+export default DeleteConfirmationModal;
--- a/frontend/src/components/intake/PermittedIndividualsForm.tsx
+++ b/frontend/src/components/intake/PermittedIndividualsForm.tsx
@@ -1,125 +1,109 @@
-import React from "react";
+import React, { useState } from "react";
 import {
   Box,
   Button,
-  FormControl,
-  FormLabel,
-  SimpleGrid,
+  Center,
+  Icon,
+  Text,
+  useDisclosure,
 } from "@chakra-ui/react";
-import { Field, Form, Formik } from "formik";
-import CustomInput from "../common/CustomInput";
-import { CustomSelectField } from "./CustomSelectField";
+import { UserPlus } from "react-feather";
+import PermittedIndividualsModal, {
+  PermittedIndividuals,
+  PermittedIndividualsDetails,
+} from "./PermittedIndividualsModal";
+import PromptBox from "./PromptBox";
 
-export type PermittedIndividualDetails = {
-  firstName: string;
-  lastName: string;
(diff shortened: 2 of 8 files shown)


frontend/src/components/intake/DeleteConfirmationModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const DeleteConfirmationModal = ({
+  isOpen,
+  onClick,
+  onClose,
+  providerName,
+}: DeleteConfirmationProps): React.ReactElement => {
+  return (
+    <Box>
+      <ModalComponent
```
bowenzhu1, line 20: is there a way to make this component more similar to the designs? i.e. smaller box, text fills entire width
image

frontend/src/components/intake/PermittedIndividualsForm.tsx
```diff
@@ -1,125 +1,109 @@
...
+        providerName: permittedIndividual.providerName,
+        phoneNo: permittedIndividual.phoneNo,
+        relationshipToChild: permittedIndividual.relationshipToChild,
+        additionalNotes: permittedIndividual.additionalNotes,
+      };
+      return individualDetail;
+    },
+  );
 
+  if (permittedIndividuals.length === 0) {
```
bowenzhu1, line 63: a slightly cleaner way instead of having two return statements doing `if ... { return }`, would be using a ternary (ex. `return ( {condition ? <first return stuff/> : <second return stuff/>})`

and that way, I think you'd only need one instance of PermittedIndividualsModal as well

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -73,30 +79,34 @@ const Intake = (): React.ReactElement => {
...
       case IntakeSteps.PROGRAM_DETAILS:
         return (
-          <ProgramForm
-            programDetails={programDetails}
-            setProgramDetails={setProgramDetails}
-            nextStep={nextStep}
-            setStep={setStep}
-          />
+          <>
+            <ProgramForm
```
bowenzhu1, line 83: is there a specific reason we want to have these as two components here?

I think we could just call PermittedIndividualsForm inside of ProgramForm? which would be better

frontend/src/components/intake/PromptBox.tsx
```diff
@@ -18,6 +28,8 @@ export type PromptBoxProps = {
   secondaryButtonIcon?: ReactElement;
   secondaryOnButtonClick?: () => void;
   individualDetails?: IndividualDetailsOverview[];
+  permittedIndividualDetails?: PermittedIndividualsDetails[];
+  editIndividual?: (index: number) => void;
```
bowenzhu1, line 32: doesn't look like we're passing in this prop anywhere in this PR?

frontend/src/components/intake/PromptBox.tsx
```diff
@@ -31,9 +43,99 @@ const PromptBox = ({
...
 }: PromptBoxProps): React.ReactElement => {
   const history = useHistory();
+  const [deleteIndex, setDeleteIndex] = useState(0);
+  const {
+    onOpen: onOpenDeletePermittedIndividuals,
+    isOpen: isOpenDeletePermittedIndividuals,
+    onClose: onCloseDeletePermittedIndividuals,
+  } = useDisclosure();
+
+  if (permittedIndividualDetails && permittedIndividualDetails.length > 0) {
```
bowenzhu1, line 58: this is a lot of repeat code - and it looks like the only major difference is the deletion confirmation modal it seems?

I think instead of having an entire separate if, return statement w repeat code, we can just add another prop along the lines of `showDeleteModal` and then just add checks/ternaries to existing components like the delete icon onClick

bowenzhu1 (changes requested):
left some comments, lmk if you have questions.

hows the other part of this going btw (view and edit details?) lmk if you're stuck there, but also 
just a heads up that you might end up with some merge conflicts with eugene's PR for the similar view and edit details functionality depending on if you try to implement it a different way

frontend/src/components/intake/DeleteConfirmationModal.tsx
xiongjasmine: edited `ModalComponent.tsx` for UI changes

bowenzhu1 (approved):
looks good! amazing work yay

PR #161, opened 2023-03-23
Add “Submit case” flow

xiongjasmine:
## Notion ticket link

[Add “Submit case” flow](https://www.notion.so/uwblueprintexecs/Add-Submit-case-flow-d3d6b0577a2e4f108533bcbadffba919)

## Implementation description
* Added 'Intake Form Complete' page and modals
* 'Submit Error' modal looks like this:
Screen Shot 2023-03-22 at 8 54 06 PM

## Steps to test
1. Go to `/intake`, go to case review page, go through submission process

* Make sure the flow is correct
* Not sure about the code style, it might be a bit messy

## Checklist


frontend/src/components/pages/IntakePage.tsx
```diff
@@ -103,6 +114,24 @@ const Intake = (): React.ReactElement => {
...
+            <Center>
+              <Image src={Logo} alt="CAS Logo" marginBottom="45px" />
+            </Center>
+            <Heading size="xl">
+              Thank you for completing the intake form
+            </Heading>
+            <Text>The admin team will review it promptly.</Text>
+            <IntakeFooter
+              nextButtonText="Return to dashboard"
+              isStepComplete={() => true} // TODO: validate form
```
bowenzhu1, line 129: do we need this TODO comment here

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -122,46 +151,61 @@ const Intake = (): React.ReactElement => {
...
+          <IntakeHeader
+            primaryTitle="Intake Form Complete"
+            secondaryTitle="Initiate New Case"
+          />
+        );
+      default:
+        return (
+          <>
+            {reviewHeader ? (
+              <IntakeHeader
```
bowenzhu1, line 174: similar comment to your other pr: we want to avoid duplicating component calls/code, and instead we can do conditional passes to the prop (primaryTitle in this case)

frontend/src/components/intake/IntakeFormFooter.tsx
```diff
@@ -84,12 +94,28 @@ const IntakeFooter = ({
...
+            history.goBack();
+          } else {
+            onNextStep();
+          }
         }}
       >
         <Box pr="5px">{nextButtonText}</Box>
         <ArrowRight />
       </Button>
+
```
bowenzhu1, line 110: we could prob also just add the submit error modal logic here too. i don't rlly like having the modal logic in the footer component lol but i think this is fine if there's no better place

frontend/src/components/intake/IntakeFormFooter.tsx
```diff
@@ -84,12 +94,28 @@ const IntakeFooter = ({
         type="submit"
         isLoading={registrationLoading}
         onClick={() => {
-          onNextStep();
+          if (nextButtonText === "Submit case") {
```
bowenzhu1, line 97: i think generally string prop comparison isn't good, and we should technically try to separate this into an enum or something buuuuut this is fine cause its simple and is alrdy an existing prop/component

bowenzhu1 (changes requested):
left some minor things to change and noted some uhhh not great (but acceptable) practices

frontend/src/components/pages/IntakePage.tsx
xiongjasmine: i rebased and updated it and it still looks a bit repetitive but i tried to do it in the cleanest way i could. can you check it again and lmk if it's ok?

bowenzhu1: yep this looks good! it's still a relatively simple case so i dont think there's that much more we can do without overkilling it

bowenzhu1 (changes requested):
small comment for the useDisclosure for SubmitErrorModal, but looks good to merge otherwise

bowenzhu1 (approved):

PR #162, opened 2023-03-27
Implement “Back to case individuals” button from Add Child

xiongjasmine:
## Notion ticket link

[Implement “Back to case individuals” button from Add Child](https://www.notion.so/uwblueprintexecs/Implement-Back-to-case-individuals-button-from-Add-Child-12e06f96d38147038aa36ccb2be64171)

## Implementation description
* Removed `/intake/add-child` route and moved add child flow to the intake page

## Steps to test
1. Type stuff into intake form
2. Go to 'Individual details' page and click 'Add child'
3. Click 'Back to case individuals' and make sure it redirects to'Individual details' page
4. Make sure the stuff you entered into the form earlier is still there

* Check that flow is correct

## Checklist

### DIFF
--- a/frontend/src/components/pages/IntakePage.tsx
+++ b/frontend/src/components/pages/IntakePage.tsx
@@ -135,19 +135,12 @@ const Intake = (): React.ReactElement => {
         );
       default:
         return (
-          <>
-            {reviewHeader ? (
-              <IntakeHeader
-                primaryTitle="Edit Case Intake Submission"
-                secondaryTitle="Case Management"
-              />
-            ) : (
-              <IntakeHeader
-                primaryTitle="Initiate New Case"
-                secondaryTitle="Case Management"
-              />
-            )}
-          </>
+          <IntakeHeader
+            primaryTitle={
+              reviewHeader ? "Edit Case Intake Submission" : "Initiate New Case"
+            }
+            secondaryTitle="Case Management"
+          />
         );
     }
   };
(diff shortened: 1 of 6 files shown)


frontend/src/components/pages/IntakePage.tsx
```diff
@@ -122,67 +124,80 @@ const Intake = (): React.ReactElement => {
...
-          maxWidth="container.xl"
-          padding="30px 96px"
-          marginBottom="100px"
-        >
-          {step !== IntakeSteps.REVIEW_CASE_DETAILS && (
-            <Button
-              leftIcon={<ArrowLeft />}
-              marginBottom="30px"
+            <UnsavedProgressModal
+              isOpen={isOpenUnsavedProgress}
```
bowenzhu1, line 192: We should try to combine this and only call `<UnsavedProgressModal>` once. 

I think we can conditionally pass in the prop values ex: `reviewVersion={reviewHeader ? reviewHeader : null},` instead of conditionally rendering two different versions of the same component

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -122,67 +124,80 @@ const Intake = (): React.ReactElement => {
...
+              <IntakeHeader
+                primaryTitle="Edit Case Intake Submission"
+                secondaryTitle="Case Management"
+              />
+            ) : (
+              <IntakeHeader
+                primaryTitle="Initiate New Case"
+                secondaryTitle="Case Management"
+              />
+            )}
```
bowenzhu1, line 149: same comment as the one for <UnsavedProgressModal>, generally I think we want to avoid duplicating component calls/code, and instead we can do conditional passes to the props

bowenzhu1 (changes requested):
left a couple comments for code style, but overall looks good!

bowenzhu1 (approved):
approving this for now so you can merge it as soon as you get your changes done - but pls do the changes first :') and lmk if u have any questions

PR #171, opened 2023-04-27
Integrate Add Child flow and child data saving

Ezzhingy:
## Notion ticket link

[Integrate Add Child flow and child data saving](https://www.notion.so/uwblueprintexecs/Integrate-Add-Child-flow-and-child-data-saving-8bbfd18fd7674c7eb166076b4e88c8d9)

## Implementation description
* Created `ChildrenDetails` type, stores child data
* Once user clicks on `Save Child Information`, `ChildrenDetails` gets updated

## Steps to test
1. `console.log` spam

* Make sure `ChildrenDetails` gets updated

## Checklist

### DIFF
--- a/frontend/src/components/intake/IndividualDetailsEntry.tsx
+++ b/frontend/src/components/intake/IndividualDetailsEntry.tsx
@@ -1,8 +1,10 @@
 import React from "react";
 import { VStack, Icon } from "@chakra-ui/react";
 import { UserPlus } from "react-feather";
-import PromptBox from "./PromptBox";
+import PromptBox, { IndividualDetailsOverview } from "./PromptBox";
 import { Caregivers } from "./NewCaregiverModal";
+// eslint-disable-next-line import/no-cycle
+import { Childrens } from "../pages/IntakePage";
 import Stepper from "./Stepper";
 import IntakeSteps from "./intakeSteps";
 import IntakeFooter from "./IntakeFormFooter";
--- a/frontend/src/components/intake/IndividualDetailsEntry.tsx
+++ b/frontend/src/components/intake/IndividualDetailsEntry.tsx
@@ -1,10 +1,8 @@
 import React from "react";
 import { VStack, Icon } from "@chakra-ui/react";
 import { UserPlus } from "react-feather";
-import PromptBox, { IndividualDetailsOverview } from "./PromptBox";
+import PromptBox from "./PromptBox";
 import { Caregivers } from "./NewCaregiverModal";
-// eslint-disable-next-line import/no-cycle
-import { Childrens } from "../pages/IntakePage";
 import Stepper from "./Stepper";
 import IntakeSteps from "./intakeSteps";
 import IntakeFooter from "./IntakeFormFooter";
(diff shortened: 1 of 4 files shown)


Ezzhingy:
*lint code breaks cause i never call the state

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -19,15 +19,26 @@ import ReferralForm, { ReferralDetails } from "../intake/ReferralForm";
...
 import { Providers } from "../intake/NewProviderModal";
+import { ChildDetails } from "../intake/child-information/ChildInformationForm";
+import { SchoolDetails } from "../intake/child-information/SchoolDaycareForm";
+
+type ChildrenDetails = {
+  childDetails: ChildDetails;
+  schoolDetails: SchoolDetails;
+  providers: Providers;
+};
+export type Childrens = ChildrenDetails[];
```
bowenzhu1, line 41: can we define all this in `AddChildPage` instead? I think that would fix a lot of the cyclical import issues

bowenzhu1: also can we use `Children` instead of `Childrens` in all cases (childrens feels too wrong LOL)

frontend/src/components/intake/IndividualDetailsEntry.tsx
bowenzhu1, line 1: is there a reason we're importing these here for this PR? If not then we can remove these imports from this PR, and save these changes for when we need them

bowenzhu1 (changes requested):
some changes requested relating to the eslint issues, but overall looks good to me

also a general suggestion is that eslint warnings shouldn't rlly be ignored - there are cases where u can do `// eslint-disable-next-line` but usually it's better to avoid breaking eslint

Ezzhingy:
i kept one instance of "childrens", cause i cant pass "children" in as a prop `children: children` doesn't work

frontend/src/components/intake/child-information/AddChildPage.tsx
```diff
@@ -53,6 +64,15 @@ const AddChild = ({
...
     // TODO: Do something with the information
+    setChildren((prevChildren) => [
+      ...prevChildren,
+      {
+        childDetails: { ...childDetails },
+        schoolDetails: { ...schoolDetails },
+        providers: [...providers],
+      },
+    ]);
+    console.log(childrens);
```
bowenzhu1, line 75: can we remove this console log

frontend/src/components/pages/IntakePage.tsx
```diff
@@ -190,6 +191,8 @@ const Intake = (): React.ReactElement => {
           allProviders={allProviders}
           setAllProviders={setAllProviders}
           setStep={setStep}
+          childrens={children}
```
bowenzhu1, line 194: wait i think `children={children}` should work? can we change the field name to `children` then? 

what error is it giving u if it's `children={children}`

bowenzhu1 (commented):
small follow ups (I'll look to approve and merge this in right after this)

bowenzhu1 (approved):

PR #176, opened 2023-05-28
Permanent Deletion Popup Modal

jacque1ine:
## Notion ticket link

[Permanent Deletion Popup Modal](https://www.notion.so/uwblueprintexecs/Permanent-Deletion-Popup-Modal-48e7696c5b7e4ee58b6486912a244cdc?pvs=4)

## Implementation description
* Made changes to button styling to add warning class so that primary button would be red
* Created new component from ModalComponent 
* Added optional fields for ModalComponent to allow to change button color and text color
* Added functionality to move state up from custom input component

## Steps to test
1. Click on Permanently Delete Button on HomePage 
image
2. In the input field, type "Confirm Deletion", this should turn the confirm button from a disabled state to an active state.
image
 

* The modal button turning enabled with Confirm Deletion is typed in.
* If changes to ModalComponent and CustomInput components causes any other bugs.

## Checklist


frontend/src/components/pages/HomePage.tsx
```diff
@@ -1,15 +1,26 @@
...
 import IntakeHeader from "../intake/IntakeHeader";
+import PermanentDeleteModal from "../dashboard/PermanentDeleteModal";
 
 const SecondaryHeader = (): React.ReactElement => {
   const history = useHistory();
   function goToIntake() {
     history.push("/intake");
   }
+
+  const { isOpen, onOpen, onClose } = useDisclosure();
```
bowenzhu1, line 23: this should be a bit more descriptive/custom (makes it better for if we add more modals here in the future), i.e. `isOpenPermanentDelete, onOpenPermanentDelete` etc.

frontend/src/components/pages/HomePage.tsx
```diff
@@ -32,6 +43,23 @@ const SecondaryHeader = (): React.ReactElement => {
         >
           New case
         </Button>
+
+        {/* //Remove this later - just for testing modal */}
```
bowenzhu1, line 47: can we add `//TODO` at the start of this comment - easier to search for all the TODOs that way

frontend/src/components/pages/HomePage.tsx
```diff
@@ -32,6 +43,23 @@ const SecondaryHeader = (): React.ReactElement => {
...
+          rounded="lg"
+          border="1px"
+          onClick={onOpen}
+          // leftIcon={<Icon as={FilePlus} />}
+        >
+          Permanently Delete
+        </Button>
+        <PermanentDeleteModal
+          isOpen={isOpen}
+          onClick={() => {}}
```
bowenzhu1, line 60: add //TODO here for adding delete action

frontend/src/components/pages/HomePage.tsx
bowenzhu1: similar to other files like ChildProviderForm

`  const {
    onOpen: onOpenNewProviders,
    isOpen: isOpenNewProviders,
    onClose: onCloseNewProviders,
  } = useDisclosure();`

frontend/src/components/common/ModalComponent.tsx
```diff
@@ -34,6 +36,8 @@ const ModalComponent = ({
   isOpen,
   onClose,
   unsavedProgressModal,
+  color,
```
bowenzhu1, line 39: we can make these prop names a bit more in line w their intended use cases for clarity:
`titleColor` and `mainButtonVariant`

frontend/src/components/dashboard/PermanentDeleteModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+}: SubmitCaseProps): React.ReactElement => {
+  const [confirmationInput, setConfirmationInput] = useState("");
+
+  const handleConfirmationInputChange = (
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    setConfirmationInput(event.target.value);
+  };
+
+  const isConfirmationValid = confirmationInput.trim() === "Confirm Deletion";
```
bowenzhu1, line 25: slightly better way of checking if it matches is to define this string in a separate const like `const deleteConfirmationText`, and then call it both here and in the prompt `Please confirm that you want to delete this case by typing {deleteConfirmationText or something here} in the text box below.`

frontend/src/components/dashboard/PermanentDeleteModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+              name="confirmDeletion"
+              type="string"
+              placeholder="Insert text here"
+              value={confirmationInput}
+              onChange={handleConfirmationInputChange}
+            />
+          </Box>
+        }
+        onClick={() => {
+          if (isConfirmationValid) {
```
bowenzhu1, line 56: if statement seems redundant here since we're already disabling the button for not valid confirmation

frontend/src/components/common/CustomInput.tsx
bowenzhu1, line 1: I dont think we need to make changes to this component at all - the useState and handleChange here seem unnecessary here since they should be already defined in whatever parent component that calls `CustomInput`

bowenzhu1 (changes requested):
overall great work! left some comments in terms of cleaning the code up, but the general implementation is solid

bowenzhu1 (approved):
lgtm! great work! (i assigned a new ticket relating to this one to you in case you didn't see)

PR #177, opened 2023-06-02
add sections for dashboard categories

kevinchen220:
## Notion ticket link

[Main Dashboard Categories](https://www.notion.so/uwblueprintexecs/Main-Dashboard-Case-Cards-819164235311446e843fa6b76fcb3b9b)

## Implementation description
* Create a component for the sections
* Pass in status name and cases to display

## Steps to test
1. See changes on the homepage dashboard

* How should the four sections be displayed?
  * It is currently calling the component 4 times

## Checklist


frontend/src/types/CaseTypes.ts
bowenzhu1, line 1: nit: name of the file + enum should be the same if possible, i.e. both `CaseStatuses` instead of CaseTypes

frontend/src/components/dashboard/FilteredSection.tsx
```diff
@@ -0,0 +1,63 @@
...
+            <Text textStyle="text-medium">
+              No current {status.toLowerCase()} cases
+            </Text>
+          </Center>
+        ) : (
+          <Flex justifyContent="space-between">
+            {cases.slice(0, 4).map((caseData: CaseCardProps) => {
+              return (
+                <CaseCard
+                  key={caseData.caseTitle}
```
bowenzhu1, line 47: we don't have case ids for now so this is fine, but note that generally the keys for mapped objects should be unique

bowenzhu1 (approved):
this looks good to me! nice work - thanks for integrating the case cards and adding dummy data

we might end up changing how the sizing/spacing of the cards works down the line, but this is good to merge now.

frontend/src/components/dashboard/FilteredSection.tsx
kevinchen220: Yup, I just thought it would be the most unique out of the existing fields

bowenzhu1 (approved):

PR #180, opened 2023-06-08
KevinC/ add logout modal

kevinchen220:
## Notion ticket link

[Implement Log Out Modal and Flow](https://www.notion.so/uwblueprintexecs/Implement-Log-Out-Modal-and-Flow-0119c7c9c45b4d57a82db4493d04b6fb?pvs=4)

## Implementation description
* Added a Log out modal using `ModalComponent`
* Log out modal is controlled by the `Logout` button
* Moved to previous logout function to the confirmation button onClick in the modal
  * Not exactly sure how to test that

## Steps to test
1. Click logout button on the top right of the dashboard
2. See the Logout Modal that pops up

* Whether the logout function is correct

## Checklist

image

### DIFF
--- a/frontend/src/components/auth/Logout.tsx
+++ b/frontend/src/components/auth/Logout.tsx
@@ -1,5 +1,5 @@
-import { Button, Icon } from "@chakra-ui/react";
-import React, { useContext, useState } from "react";
+import { Button, Icon, useDisclosure } from "@chakra-ui/react";
+import React, { useContext } from "react";
 import { LogOut } from "react-feather";
 
 import authAPIClient from "../../APIClients/AuthAPIClient";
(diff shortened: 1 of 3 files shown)


frontend/src/components/common/ModalComponent.tsx
```diff
@@ -46,6 +47,7 @@ const ModalComponent = ({
   onLeftButtonClick,
   titleColor,
   mainButtonVariant,
+  showModalCloseButton = true,
```
jess94yng, line 50: Currently when the log out modal first pops open, the cancel button is in a selected state (blue ring around it) - see if that can be fixed?

jess94yng (changes requested):

frontend/src/components/auth/Logout.tsx
```diff
@@ -1,33 +1,47 @@
...
+import React, { useContext, useState } from "react";
 import { LogOut } from "react-feather";
 
 import authAPIClient from "../../APIClients/AuthAPIClient";
 import AuthContext from "../../contexts/AuthContext";
+import LogoutModal from "./LogoutModal";
 
 const LogoutButton = (): React.ReactElement => {
   const { authenticatedUser, setAuthenticatedUser } = useContext(AuthContext);
+  const [openModal, setOpenModal] = useState(false);
```
bowenzhu1, line 11: we could use `useDisclosure` instead here which would be a bit cleaner. - to make use of ChakraUI modal handling logic. you can cmd+shift+f search for other files where we use it for reference. but other than that lgtm

frontend/src/components/common/ModalComponent.tsx
kevinchen220: Looks like that's the default behaviour of ChakraUI modals, the first tab-able element is selected. Not sure if there is an easy fix

kevinchen220: Nevermind I got it to work

bowenzhu1 (approved):
lgtm to merge

jess94yng (approved):

PR #187, opened 2023-06-25
Case overview skeleton pr changes

odis43:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Case-Overview-page-skeleton-a45ba23e6f15411f9c1012dc10f3588b?pvs=4)

## Implementation description
* Changes to PR review by Bowen and Jessica. Changes include fixing the design to meet the requirements of the figma document, improving code quality through the use of component based styling and removing unnecessary logic from the footer (caseOverviewFooter)

## Steps to test
1. Check figma documentation for design match up
2. Check code for mistakes in implementation 

* If the design is more suitable as per the figma document

## Checklist
- [ x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x ] I have run the appropriate linter(s)
- [ x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


frontend/src/components/intake/CaseOverviewFooter.tsx
```diff
@@ -0,0 +1,57 @@
...
+      justify="flex-end"
+      flexWrap="wrap"
+      padding="20px"
+      left="0"
+      position="fixed"
+      bottom="0"
+      zIndex="5"
+    >
+      <Button
+        height="38px"
```
bowenzhu1, line 30: do we need to specify the dimensions here? I think the default dimensions should work, which we use for other instances of the button if im not mistaken. lmk if there's a reason for specifying this though

frontend/src/components/intake/CaseOverviewFooter.tsx
```diff
@@ -0,0 +1,57 @@
...
+        onClick={() => {
+          // TODO: Handle generate report logic
+        }}
+      >
+        <div style={{ paddingRight: "10px" }}>
```
bowenzhu1, line 35: having div wrappers like this is generally not that great of a practice, most components should be able to have CSS applied (like padding) directly onto them

I think the FileText is just a icon though, but in a case like this you can use the <Icon> component and pass in FileText

frontend/src/components/intake/CaseOverviewFooter.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <FileText width="13px" />
+        </div>
+        Generate Report (PDF)
+      </Button>
+      <Box flex="1" /> {/* Spacer element */}
```
bowenzhu1, line 40: in general try to avoid adding components if you dont have to, to reduce clutter/improve code cleanliness.

 i think you can wrap the two buttons in a Box or something and then apply the spacing there.

frontend/src/components/intake/CaseOverviewFooter.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <Button
+        width={{ sm: "95vw", md: "45vw", lg: "auto" }}
+        height="38px"
+        background={colors.red[600]}
```
bowenzhu1, line 44: might be able to use an existing buttonStyle instead of applying the color directly

frontend/src/components/pages/CaseOverview.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+const SecondaryHeader = (): React.ReactElement => {
+  return (
+    <Box>
+      <Button variant="ghost">
+        <ArrowLeft style={{ width: "50px", height: "50px" }} />
+      </Button>
+
+      <Flex pt="10" mt="-7">
+        <div>
```
bowenzhu1, line 16: this div is unnecessary

frontend/src/components/pages/CaseOverview.tsx
```diff
@@ -0,0 +1,265 @@
...
+import colors from "../../theme/colors";
+
+const SecondaryHeader = (): React.ReactElement => {
+  return (
+    <Box>
+      <Button variant="ghost">
+        <ArrowLeft style={{ width: "50px", height: "50px" }} />
+      </Button>
+
+      <Flex pt="10" mt="-7">
```
bowenzhu1, line 15: i think negative margins is generally not a good practice or at least not typically used, is there another way we can format this

frontend/src/components/pages/CaseOverview.tsx
```diff
@@ -0,0 +1,265 @@
...
+        <div>
+          <b color={colors.black.default} style={{ fontSize: "12px" }}>
```
bowenzhu1, line 17: we should use the <Text> component since we're using ChakraUI. Using framework components helps make things consistent instead of using native html

frontend/src/components/pages/CaseOverview.tsx
```diff
@@ -0,0 +1,265 @@
...
+            CHILDREN SERVICES OR KINSHIP WORKER
+          </b>
+          <div
```
bowenzhu1, line 20: similar comment here to use framework components instead of native <divs> , we should use <Box> or <Flex> or something instead

frontend/src/components/pages/CaseOverview.tsx
```diff
@@ -0,0 +1,265 @@
...
+              borderColor="#8397c0"
+              variant="outline"
+            >
+              Save
+            </Button>
+          </div>
+        </div>
+      </Flex>
+
+      <div style={{ paddingBottom: "100px" }}>
```
bowenzhu1, line 40: i dont think we need this div either, could probbably apply the padding to the Flex component right below this

frontend/src/components/pages/CaseOverview.tsx
```diff
@@ -0,0 +1,265 @@
...
+        <Flex direction="row" justifyContent="space-between" gap="10">
+          <Flex direction="column" flex="1">
+            <Flex pt="50">
+              <h1 style={{ fontSize: "23px", fontWeight: 600 }}>Children</h1>
```
bowenzhu1, line 44: should use Text or Header here as well

frontend/src/components/pages/CaseOverview.tsx
```diff
@@ -0,0 +1,265 @@
...
+              p="4"
+              mt="4"
+              h="250px"
+              position="relative"
+              display="flex"
+              alignItems="center"
+              justifyContent="center"
+              flexDirection="column"
+            >
+              <b style={{ fontSize: "20px", fontWeight: 500 }}>
```
bowenzhu1, line 60: similar comment here as the above, and should be applied for the rest of the file 
- use Framework components instead of native <divs> and html components like <b> <p>
- try to avoid having unnecessary/wrapper components/divs
- use existing styles for components if you can for consistency instead of manually defining CSS

bowenzhu1 (changes requested):

frontend/src/components/intake/CaseOverviewFooter.tsx
odis43: done

frontend/src/components/intake/CaseOverviewFooter.tsx
odis43: hmm, i have the padding there to have between the icon and the text Generate Report, removing it brings the icon and the text together with no whitespace

frontend/src/components/intake/CaseOverviewFooter.tsx
odis43: resolved

frontend/src/components/pages/CaseOverview.tsx
odis43: resolved

frontend/src/components/pages/CaseOverview.tsx
odis43: removing the div changes the formatting, should i still remove?

frontend/src/components/pages/CaseOverview.tsx
odis43: resolved

frontend/src/components/pages/CaseOverview.tsx
odis43: resolved

frontend/src/components/pages/CaseOverview.tsx
odis43: resolved

frontend/src/components/pages/CaseOverview.tsx
odis43: resolved

bowenzhu1 (approved):

PR #196, opened 2023-07-07
Sohail/view all cases frame

SohailSayed:
## Notion ticket link

[Main Dashboard View All Cases Frame](https://www.notion.so/uwblueprintexecs/Main-Dashboard-View-All-Cases-Frame-59c9aa4ca2a04b20a1358ad7bf9793e7)

## Implementation description
* Create new cases page, display grid of specific type of cases, filtered by query params in URL
* Organize and display case cards in new FilteredCaseDisplay component
* Use empty "filler" box to occupy space when row has less than 4 cases

## Steps to test
1. Direct yourself to the cases page by either clicking on one of the "View All" buttons on Homepage, or typing the URL directly (example http://localhost:3000/cases?status=active)

image
image

2. Cases should be centered on the page in a grid with 4 cards per row, with the header and back arrow aligned with the leftmost case

* Ensuring spacing issues don't pop up (especially on smaller laptop screen sizes)
* I'd like a second opinion on my current implementation for when there are no cases (can be viewed at http://localhost:3000/cases?status=submitted)

## Checklist


frontend/src/components/common/FilteredCaseDisplay.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <Center height="full">
+        <Text textStyle="text-medium">
+          No current {status.toLowerCase()} cases
+        </Text>
+      </Center>
+    );
+  }
+
+  const rows = [];
+  for (let i = 0; i < numberOfRows; i += 1) {
```
bowenzhu1, line 27: this works, but it's a pretty hacky solution. Maybe if we set the Flex justifyContent to be flex-start, then we don't need the additional filler boxes

frontend/src/components/pages/CasesPage.tsx
```diff
@@ -0,0 +1,165 @@
...
+        {formattedStatus} Cases
+      </Text>
+      <Box pt="32px">
+        <FilteredCaseDisplay
+          cases={cases[status]}
+          numberOfRows={2}
+          status={status}
+        />
+      </Box>
+      <Flex justifyContent="center" py="45px">
```
bowenzhu1, line 148: shouldnt show this if cases[status] is empty, could conditionally check that

bowenzhu1 (changes requested):
For the empty state, we shouldn't show the load more button.

Also, not a hard requirement, but could we try making it so that the container wraps when the window is super small, instead of overflowing off the screen. We might be able to use a <Flex> in CasePage to wrap the FilteredCaseSections for this. Screenshot 2023-07-09 at 11 27 28 AM

on the flipside, the container seems to spill if the window is really big as well: 
Screenshot 2023-07-09 at 11 29 18 AM

frontend/src/components/pages/CasesPage.tsx
SohailSayed: Just pushed some changes:



> shouldnt show this if cases[status] is empty, could conditionally check that

Load more cases button doesn't show when cases[status] is empty now

> this works, but it's a pretty hacky solution. Maybe if we set the Flex justifyContent to be flex-start, then we don't need the additional filler boxes

No longer using empty case cards/fillers 

>on the flipside, the container seems to spill if the window is really big as well:

This should also be fixed

>Also, not a hard requirement, but could we try making it so that the container wraps when the window is super small, instead of overflowing off the screen. We might be able to use a in CasePage to wrap the FilteredCaseSections for this.

It'll be hard to implement wrapping on the cards since each row (which can have up to 4 cards) is a separate flex container. I could try to implement it so that if the screen size is below a certain size, it switches from showing 4 cards per row to 2 cards per row. Is that worth trying?

bowenzhu1 (approved):

PR #199, opened 2023-07-14
integrate dashboard components and flows

bowenzhu1:
## Notion ticket link
no ticket

integrating a lot of different tickets/components together

## Implementation description
* active case card clicks into Case Overview
* visit cadence and archive case modals hooked up to Case Overview page
* update db seed with more cases
* revert IntakeAPIClient to previous commit from #198 , to map out entire intake response
* some refactoring of Case Status types 

## Steps to test
1. the app should be more functional now and more flows are accessible
2. The preview link doesn't really work with the backend right now. You can pull the branch and then run docker commands (wipe docker/dbs, compose, upgrade, seed, etc.) to have the app get cases from the backend

* 

## Checklist

### DIFF
--- a/frontend/src/components/pages/HomePage.tsx
+++ b/frontend/src/components/pages/HomePage.tsx
@@ -75,7 +75,7 @@ const Home = (): React.ReactElement => {
         caseLead: intake.caseReferral.referringWorkerName,
         date: intake.caseReferral.referralDate,
         familyName: intake.caseReferral.familyName,
-        caseTag: intake.intakeStatus as CaseStatus,
+        caseTag: intake.intakeStatus,
       }));
     }
     return [];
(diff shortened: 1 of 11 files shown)


frontend/src/components/pages/HomePage.tsx
```diff
@@ -59,39 +47,17 @@ const SecondaryHeader = (): React.ReactElement => {
...
       </Flex>
     </Box>
   );
 };
 
 const Home = (): React.ReactElement => {
+  // TODO: remove console log and use context instead of state
+  const casesFromContext = useContext(CasesContext);
+  // eslint-disable-next-line
+  console.log(casesFromContext);
```
jacque1ine, line 59: Nit: Should there be a console log here?

frontend/src/types/CasesContextTypes.ts
```diff
@@ -1,6 +1,7 @@
 export type Case = {
   user_id: string | number;
   case_id: string | number;
+  intakeStatus: string;
```
jacque1ine, line 4: Not sure if this is an option but does it make more sense to make this an ENUM instead since there are only 4 possible statuses.

jacque1ine (approved):

jacque1ine (commented):
FYI when I clicked on the preview link I still don't see the cases from the db in the dashboard page. I just see empty cards. Am I suppose to be able to see the cases?

frontend/src/components/pages/HomePage.tsx
bowenzhu1: yea there rlly shouldn't be, but I left it there for now and marked it as a TODO to remove since we still need to actually implement the context logic

frontend/src/types/CasesContextTypes.ts
bowenzhu1: good call out

bowenzhu1:
yea I don't think the preview can get data frmo the backend yet, need to look into that to see if it's setup incorrectly

